PR #142, opened 2023-10-23
Swagger API demo for our docs site

nrichers:

This PR demos how we could add better API reference docs to our site via Swagger and RapiDoc. It uses the backend API because it had the Swagger JSON spec readily available. 

- `_quarto.yml`: Add a new top navbar item for "Backend API"
- `backend-api.qmd`: new top-level API docs entry point that imbeds the `backend-swagger.json` via RapiDoc, including some styling tweaks

image

Nice stuff we get basically for free via this PR:

- generated examples
- sample cURL command with payload
- interactivity (needs an API key to work)

The backend Swagger JSON file is not perfect, but I made no tweaks to it for this demo. Ideally, we'd have more descriptive sidebar navigation, but this is what it looks like if you don't show the path in the navbar

image

To get this output, remove

- `show-method-in-nav-bar="as-plain-text"`
- `use-path-in-nav-bar="true"`


panchicore:
cool!

PR #145, opened 2023-11-07
November 9 release notes

nrichers:

This PR includes the release notes for November 7. Since there was not enough content to justify having multiple pages, I combined everything into one page.

## Output

Note that the test descriptions entry uses an animated GIF that this screen capture does not show.

Capto_Capture 2023-11-09_02-00-27_pm


mehdi0501 (commented):
The Release notes look good, however I don't see mention of the newly added support for embedding models and clustering models. Those are pretty value-added features that deserve to be mentioned :)

nrichers:
> The Release notes look good, however I don't see mention of the newly added support for embedding models and clustering models. Those are pretty value-added features that deserve to be mentioned :)

As per a convo I had with @cachafla in https://validmind.slack.com/archives/C03169FKW3C/p1699290194779719?thread_ts=1699058170.093839&cid=C03169FKW3C, these are not to be externalized publicly in this release (the release notes have been written but they were commented out [for the previous release](https://github.com/validmind/documentation/blob/main/site/releases/2023-oct-25/highlights.qmd#L12-L46). 

If you are OK with publishing the release, let me know? I can update the date to match whatever day we go live with this.

PR #147, opened 2023-11-17
Remove safe external dependencies

nrichers:

This is a companion PR to #146 that removes only those external dependencies that are safe to remove:

- Locally stored Google fonts 
- Unneeded external font references in our template schema docs

This PR should be merged as it reduces our dependencies and improves site loading times, as we no longer need to fetch our main font from Google every time we load the site. 

There are no output changes to review, as the site largely looks almost exactly the same, except for the template schema docs that now use the same font as the rest of our docs site.


site/guide/edit-templates.qmd
```diff
@@ -17,7 +17,7 @@ If you are not sure which template or which version of a template a documentatio
 ## Template schema
```
nrichers, line 17:
```suggestion
## Documentation template schema
```

mehdi0501 (approved):

PR #159, opened 2024-03-13
Run release notes check only if not in draft mode

nrichers:

This PR makes a small tweak to our release notes check so that it runs only if a PR is not in draft mode. This should avoid frequent CI check failures when a PR is in progress. 

Here's how the CI check gets skipped after you change the PR to draft mode and then make a commit: 

image

And here's what happens when you change the PR back to ready for review: 

image


nrichers:
This PR has been approved in other repos. Will merge this one.

PR #172, opened 2024-04-17
Update _variables.yml

validbeck:
Just popped some product variables into our YAML file!

(I used the GH website to make this pull request as I'm still a bit lost via the local way, oops)

### DIFF
--- a/site/_variables.yml
+++ b/site/_variables.yml
@@ -5,3 +5,5 @@ vm_ui: https://app.prod.validmind.ai
 vm_ui_privatelink: https://private.prod.vm.validmind.ai
 api_host: https://api.dev.vm.validmind.ai/api/v1/tracking
 contact: "[**Contact Us**](https://validmind.com/demo/)"
+vm_platform: "ValidMind Platform UI"
+vm_framework: "ValidMind Developer Framework"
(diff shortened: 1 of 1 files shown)


nrichers (approved):
LGTM! Kudos on finding your way to our Quarto variables file. 🚀 

One question that kind of also relates to the glossary and style guide discussions we're having separately: what about short forms that aren't proper nouns, do we want to turn those into variables as well? 

E.g. the short forms for the two variables you added are "developer framework"  and "platform UI". The former is used quite often, the latter somewhat rarely. 

We might also want to look into figuring out how to make it easier to use these variables with Quarto. One of the main reasons I didn't start using them for product names before is that they're a bit of a pain to use and there's no predictive typing at the moment. Perhaps having some product name snippets we make available for ourselves in VS Code might be helpful.

validbeck:
> One question that kind of also relates to the glossary and style guide discussions we're having separately: what about short forms that aren't proper nouns, do we want to turn those into variables as well?
>
> E.g. the short forms for the two variables you added are "developer framework" and "platform UI". The former is used quite often, the latter somewhat rarely.

Great point, added the short forms as well.

> We might also want to look into figuring out how to make it easier to use these variables with Quarto. One of the main reasons I didn't start using them for product names before is that they're a bit of a pain to use and there's no predictive typing at the moment. 

Not quite sure how to do this independently, so let's chat about it!

PR #176, opened 2024-04-26
Updated the About section to break down the articles into new categor…

validbeck:
## Internal Notes for Reviewers
- I called `quarto render` this time and checked the static files in `_site`, so hopefully everything goes smoothly...
- I've updated the old Style Guide in Notion by sending it to the Community archives and linking out to the new publicly hosted version 
- @nrichers, after you approve, permission to send this to prod and update the site thru AWS? Thank you!

> **Question:** How do we reindex the pages for Google search? We touched on this briefly yesterday but I can't remember if we performed that action. 

### Updated "About" section
| About Before      | About After |
| :-----------: | :-----------: |
| old_about | about_vm |
- Changed the sidebar link for **What is ValidMind?** to `About ValidMind` to match the structure of the landing page titles

#### Features
- Made some small changes to two pages based on how we talk about these features on the VM WordPress site: 
    - Adjusted the link for the automation page to **Automated testing & documentation** (left the full title on the page itself)
     - Adjusted the title for **Model risk governance management** to `Model risk management & governance`

| Old automation link      | New automation link |
| :-----------: | :-----------: |
|old_about|automation|

| Old governance page      | New governance page |
| :-----------: | :-----------: |
|old_governance|governance|
- Moved **ValidMind glossary** here, to break out the terms and style guide into their own sections (see below)

#### Contributing
community

- Added a brand spankin' new **The ValidMind community** page, complete with invitations to participate and contribute 
- Moved the "ValidMind style guide" here, and made two adjustments to the contents:
    - Added an example for highlighting UI elements under `Emphasis`
    - Added a note under `Product names` to caveat that variables will not work in .ipynb

| Emphasis      | Variables |
| :-----------: | :-----------: |
|emphasis|variables|

#### Fineprint
fineprint

- **Data privacy policy** and **Software license agreement** live here now under `Releases` 

### Purge external release notes for #174
- Please ignore the half-baked version of the external release notes on [pull request #174](https://github.com/validmind/documentation/pull/174) and go with the below version instead

## External Release Notes
about_vm

### About ValidMind
We've revamped our documentation for a cleaner, more intuitive experience, featuring a brand new **[About]((../about/overview.qmd))** section:

- Product overviews and our glossary now live here under *Features*
- *Releases* have been moved here as well
- Important documents such as our terms are under *Fineprint*

#### Contributing
Under this section you'll also find a brand new set of guides on *Contributing*, where you learn how to engage with the ValidMind community, read about our brand voice and vision, and more! 

[...]

PR #177, opened 2024-04-29
Nrichers/sc 2887/release notes script updates

nrichers:

This PR reimplements our release notes script from scratch.

What's better about this script:

- **Specify any number of release tag URLs**. The old script could handle only one release tag per repo but our pattern over the past year has been that we often combine multiple release tags into a single set of release notes. This meant running the old script several times and manually collating the files. 
- **Generate release notes as a single file**. When we created our original release notes script, we separated the different label categories into multiple output files. We don't generate that much content, though, and manually collated those releases notes into a single file. We now generate a single file from the start. 
- **Requires much less editing**. Better markup, with H3 headings for all entries instead of the previous list items is much easier to work with and requires far fewer manual tweaks to look right. For example, additional markup inside of the release notes such as additional list items now typically just work as-is. Bonus: the "On this page" right side navigation now lists the entries in each section. 
- **Better at handling exceptions**. The old script was missing quite a bit of error handling. For example, it had not been able to correctly embed the release notes into _quarto.yml for some time due to changes to navigation in the docs site, but it never returned any errors. 

Other assorted improvements include a default publishing date three business days from running the script,  doing away with templates file, and moving the script into a new scripts/ path. 

### How to test

1. `gh pr checkout 177`
2. `cd site`
3. `make release-notes`

After generating the release notes, run `quarto preview` to see the new release notes.

### Output 

New release notes output files:

image

Release notes embedded into _quarto.yml:

image


validbeck:
I just tried this out, but it looks like I have to first make a release note in GitHub — can I pull in the external release note from PR #176 to add, or is there another process?

nrichers:
> I just tried this out, but it looks like I have to first make a release note in GitHub — can I pull in the external release note from PR https://github.com/validmind/documentation/pull/176 to add, or is there another process?

You just need to feed it any number of existing full release URLs, e.g. 

- https://github.com/validmind/documentation/releases/tag/v2.0.7
- https://github.com/validmind/frontend/releases/tag/v1.19.0
- https://github.com/validmind/developer-framework/releases/tag/v2.0.7

validbeck:
> EDIT: And also try stuff that doesn't actually exist to test the error handling. 😁

Yep, I initially tried to plunk in my PR #176 and it didn't generate anything in the `release-notes.qmd` file. I tried the releases you suggested and it did compile them into a draft file! I assume we then clean these up and change it into a "Highlights" file?

- How would I include the release notes from the PR I made above?

nrichers:
>  I assume we then clean these up and change it into a "Highlights"

No, that file name is a leftover from the days when we generated a slew of different files for the different labels (highlights, bug fixes, etc.) and then manually combined them into one. The file should just be called `release-notes.qmd` to reflect what's in it. (The date is provided by the URL.)

nrichers:
> How would I include the release notes from the PR I made above?

You'd need to create a new release tag that includes the PR. Instructions forthcoming in the process I'm writing down for us.

nrichers:
BTW, your questions about release tags make me think that the new, recent default of NOT playing animated GIFs in GitHub is a problem. The animated GIF basically shows what kind of release URLs you can enter. 🙃

validbeck:
> BTW, your questions about release tags make me think that the new, recent default of NOT playing animated GIFs in GitHub is a problem. The animated GIF basically shows what kind of release URLs you can enter. 🙃

Yes, I figured it out after I watched the GIF, the first attempt was just a mistake on my part!

validbeck (approved):
It works as described — now to update our release notes!

validbeck:
Ah, looks like the script is conflicting with our new `About` section where we moved the overview, contribute, releases & fineprint into guide — how do we resolve that? 🤔

nrichers:
> Ah, looks like the script is conflicting with our new `About` section where we moved the overview, contribute, releases & fineprint into guide — how do we resolve that? 🤔

I fixed the merge conflict and retested to make sure the script still works. Previously, this would have broken the script, but as long as the embed marker for updating _quarto.yml is in the right place, the script will work.

PR #180, opened 2024-04-30
Minor updates to sandbox instructions

nrichers:

Following a demo today, I noticed a couple of things that can be improved in the sandbox instructions: 

- Make sure model names we say are registered in the model inventory match what's actually in the sandbox (see the screenshot for reference). 

   image

- Make it clear that the sandbox is not for proprietary information. This was highlighted during the demo and we should include that info in the callout we already had. 

   image


validbeck (approved):
Nice! I love the `For testing and evaluation purposes only` callout.

One question I have (and forgive me if this is a dumb question, sometimes my brain has holes) is we're directing folks to a UI link that has `prod` in the URL but we call it the sandbox — do you think this is something that can be misleading for our users?

nrichers:
> One question I have (and forgive me if this is a dumb question, sometimes my brain has holes) is we're directing folks to a UI link that has `prod` in the URL but we call it the sandbox — do you think this is something that can be misleading for our users?

That's a very good observation! What we are providing is a sandbox that emulates a production environment and it basically runs on our own `prod` site for the Platform UI. We, somewhat confusingly, also have a test sandbox that runs inside our docs site. In the future, we will also need to figure out where training will happen which will have its own sandbox environment. The term "sandbox" does seem to resonate with developers, though. 

No answers from me here, just the acknowledgement that we will need to figure out more of the terminology in the future and be consistent in how we use these terms.

nrichers:
> What is the live URL for our actual production site?

It's https://app.prod.validmind.ai/ — not sure if we would allow custom subdomains. That might be a question for the #engineering channel. 

>Maybe "UI Sandbox" and "Test (suite) sandbox"?

It's not just a UI sandbox, though, you use both Jupyter Hub (run notebooks) and use the UI. And the test sandbox likely just needs to be renamed to something like "Explore ValidMind tests" or something like that. It's a beta feature at this point and quite a bit more work is needed to make it look right.

PR #184, opened 2024-05-07
Remove closed beta info

nrichers:

With us winding down the closed beta program, we need to remove the signup from the docs site. 

This PR:

- Moves the beta content into the site-unused/ folder where we keep stuff we might want to refer to again in the future
- Removes the closed beta signup page from our docs site

No release notes for this — we plan to provide a final beta update and offer folks some swag as thanks.


validbeck (approved):
- I experienced some odd issues trying to review this PR (potentially user error 😓), but I more or less got the preview working! It looks like there are some conflicting files which may be the cause of the `Diverging branches can't be fast-forwarded` error I got. 
- It also notably flagged `CONFLICT (content): Merge conflict in site/_site/search.json` even when I managed to resolve the previous error;  presumably this is related to the fixing conflicts conversation. (Pointing this one out as it also told me "WARN: Unable to read search index - it may be corrupt. Please re-render the entire site to create a valid search index." and indeed the search no longer works in the preview...)
    - Can we maybe book a session I can record to go over conflicts and merging? 

Otherwise, it looks good!

PR #186, opened 2024-05-09
Fix .gitgnore files and add .env example file

nrichers:

This PR:

- Fixes an issue with duplicate `.gitignore` files that are causing issues with our usage of a `.env` file for OpenAI API credentials
- Adds a `.env.example` file that you can adopt for your own usage

You can test by checking out this PR, modifying the `site/.env` file and then running `git status`. The modified `.env` file should NOT show up.


validbeck:
Tried this out, the `.env` file never showed up when `git status` was called before this but `git clean -dfx` still cleans up the added `.env` file. 😬 Perhaps I'm doing something wrong?

validbeck (approved):
This change appropriately sets up an example `.env` file for the `make release-notes` script to run, but we should make sure our internal processes for "cleanup" on local branches doesn't include ignoring what's within `.gitignore`

nrichers:
> we should make sure our internal processes for "cleanup" on local branches doesn't include ignoring what's within `.gitignore`

Took a quick look for process docs that might mention the overzealous `git clean` command and didn't find any.

PR #203, opened 2024-05-24
Cleaned up more docs site variables

validbeck:
### Edge cases
Anything remaining in the site `.qmd` files that come up in searches for these are because:
- They are within an alt text or Mermaid chart, as those cannot accommodate variables*
- They were a direct quotation of something that used to display in the UI in a highlight/release file*
- They are part of a code comment, and should be human readable within the file
- For `framework`: They are referencing a different kind of framework, such as "governance framework" in regulatory guidance discussions

>*For these, I've left a comment above the instance somewhere along the lines of `<!-- Using the variable in alt text messes up the image display  -->`


site/guide/work-with-content-blocks.qmd
```diff
@@ -24,7 +24,7 @@ Test-driven block
 
 ## Add content blocks
 
-1. [Log in to the ValidMind UI]({{< var vm_url >}}).
+1. [Log in to the {{< var vm_platform >}}]({{< var vm_url >}}).
 
 2. In the left sidebar, click [**Model Inventory**]({{< var vm_url >}}/model-inventory).
```
nrichers, line 29: Not within the scope of this PR but it feels like these compound URLs with variable + URL path don't buy us anything and they are likely more prone of breakage. Let's see how these pan out in the future? If they do break, a good rule of thumb might be to say that either all of a URL should be a variable or none of it.

nrichers (approved):
LGTM! One comment about what the preferred practice should be for compound URLs (URLs that are part variable, part URL path). :shipit: 

Also, I missed this in my review of your first PR, but we broke all the ze buttons, as variables inside of HTML do not work. For example:

image

(Will be fixed in my https://github.com/validmind/documentation/tree/nrichers/sc-4580/bny-create-training-collateral branch and already fixed on docs.validmind.ai as a temporary measure.)

validbeck:
> but we broke all the ze buttons, as variables inside of HTML do not work.

Oh durr, thanks for catching that.

> Not within the scope of this PR but it feels like these compound URLs with variable + URL path don't buy us anything and they are likely more prone to breakage. 

Great point, I'll keep an eye out.

PR #204, opened 2024-05-27
Documentation for inserting math formulas

noosheenv:

I added a section to "Work with content blocks" under "Add content blocks" called "Add mathematical formulas". Feedback on the wording and level of detail is appreciated! Should I go more step-by-step, or is it best to just have that link to CK Editor's docs?
image


site/guide/work-with-content-blocks.qmd
```diff
@@ -50,6 +50,9 @@ Test-driven block
 After you have completed these steps, the new content block becomes a part of your model documentation.
 
 <!--- TO DO We might need to revisit terminology: we talk about "block", "section", and "module", but they all seem to be names for very similar things.--->
+### Add mathematical formulas
+
+While editing a content block, you can insert math equations using the editor. Its interface is accessed by clicking the square-root icon within the toolbar. You can use the interface to type out the equation, or paste in a LaTeX formula. See [the editor's docs](https://ckeditor.com/docs/ckeditor5/latest/features/math-equations.html) for detailed instructions.
```
nrichers, line 55: FYI, links to external docs can break and I don't think external links get checked when we generate docs. Seems useful here, but I would use links like these sparingly, as they invariably all break eventually.

nrichers, line 55:
```suggestion
While editing a content block, you can insert math equations using the editor. Its interface is accessed by clicking **√** in the toolbar. You can use the interface to type out the equation, or paste in a LaTeX formula. See [the editor's docs](https://ckeditor.com/docs/ckeditor5/latest/features/math-equations.html) for detailed instructions.
```

We should use icons where they are available (Quarto also supports Font Awesome though there is no exact match for this icon).

site/guide/work-with-content-blocks.qmd
```diff
@@ -50,6 +50,9 @@ Test-driven block
 After you have completed these steps, the new content block becomes a part of your model documentation.
 
 <!--- TO DO We might need to revisit terminology: we talk about "block", "section", and "module", but they all seem to be names for very similar things.--->
+### Add mathematical formulas
```
nrichers, line 53: As a general comment, tasks should have numbered steps to show me how to do something. Your text tells me that I can do something but _how do I actually add a formula_? An example or a screenshot with an animated GIF that shows how to do this would also be helpful as this feature is kinda nestled away. 

(The CK Editor docs are lacking here as well, BTW, though they do have a demo of sorts.)

nrichers (changes requested):
Great start for these docs! My main comment would be that this is a task which should be procedural, i.e. it should have numbered steps.

site/guide/work-with-content-blocks.qmd
```diff
@@ -50,6 +50,19 @@ Test-driven block
...
+
+While editing a content block, you can insert math equations using the editor. 
+
+#### Steps
```
nrichers, line 57: Since the parent is already at the H3 level, you could just omit this steps heading and end line 55 in a `:`.

site/guide/work-with-content-blocks.qmd
```diff
@@ -50,6 +50,19 @@ Test-driven block
...
+
+1. Click the **√** in the toolbar while editing a content block.
+
+2. You can use the interface to type out the equation, or paste in a LaTeX formula:
+
+![](mathtype-full-demo.gif){width=100%}
```
nrichers, line 63:
```suggestion
![](mathtype-full-demo.gif){width=100% fig-alt="An animation that shows how to insert a mathematical formula using LaTex."
```

For accessibility reasons, we should include some alt text.

nrichers (approved):
A couple of small additional comments on the new content but LGTM otherwise (looks GREAT to me). 🚀🚀🚀

validbeck (approved):
Nice, looks good to me as well. :)

PR #205, opened 2024-05-28
Adding model vendor support

noosheenv:
### I added steps to `guide/register-models-in-model-inventory.qmd` to document the new feature:

| Before | After |
| ----- | ----- |
| image | image |

### I added the term "model vendor" to the glossary:
image

*Should we also add "vendor model"

### I added a callout in `guide/supported-models.qmd` under "Supported model types" to show that we support vendor models:
image


validbeck (changes requested):
> *Should we also add "vendor model"

- Yes, and make sure it's a **span with an ID** so we can link to the anchor on the page! 
- You'll need to dig into your HTML knowledge here, or take a look at the code showcased in the examples for help ;) (I noticed that the link below is actually broken, it should be `https://docs.validmind.ai/about/glossary/glossary.html#platform-model-inventory` in full, so you'll want to fix that while you're there as well!

| Example anchor link | Example span ID |
|---|---|
|Screenshot 2024-05-28 at 2 54 43 PM|Screenshot 2024-05-28 at 2 55 07 PM|

> I added steps to guide/register-models-in-model-inventory.qmd to document the new feature

I'd make the first `vendor model` a link to your anchor once it's available.

> I added a callout in guide/supported-models.qmd under "Supported model types" to show that we support vendor models

I'd make `third-party vendor models` a link to the anchor here as well once it's available.

validbeck:
@noosheenv 

> The two spots that link to vendor-model won't work until it's deployed on prod because it's linking to the website instead of a file, hope that's okay!

You can spot check it in the static rendered `_site` files ;)

validbeck:
@noosheenv I just took a look at your files, and I see what you meant by not being able to double-check the links. The format should actually follow the "good" structure, i.e. (instead of linking out to the live website):

```
../../about/glossary/glossary.html#platform-model-inventory
```

Are you able to adjust the links you added to the above, and that way you can double-check and make sure the spans work? I also noticed your formatting for the spans is a little wonky, and I'm not sure they will perform in the same way with your spacing/wrapping of the text. Can you adjust them to look like the following:

```
model inventory <span id="platform-model-inventory">
```

site/guide/register-models-in-model-inventory.qmd
```diff
@@ -26,7 +26,8 @@ To be able to document models using ValidMind's model documentation and validati
    - Select the relevant model methodology being used
    - Select the relevant model use case
    - Provide a purpose statement to explain what the model will be used for
-   - select the preliminary risk tier for the model
+   - Select the preliminary risk tier for the model
+   - If the model is a [vendor model](https://docs.validmind.ai/about/glossary/glossary.html#vendor-model), toggle **Is Vendor Model** and provide the vendor name.
```
nrichers, line 30:
```suggestion
   - If the model is a [vendor model](../about/glossary/glossary.qmd#vendor-model), toggle **Is Vendor Model** and provide the vendor name.
```

Use a relative link. This works, I tested it, though we do have an issue with the navbar at the top covering the text.

site/guide/supported-models.qmd
```diff
@@ -53,6 +53,10 @@ Large language models (LLMs)
 : - Classification: Categorizes input into predefined classes.
 - Text summarization: Generates concise summaries from longer texts.
 
+:::{.callout-tip}
+ValidMind offers support for both first-party models and [third-party vendor models](https://docs.validmind.ai/about/glossary/glossary.html#vendor-model).
```
nrichers, line 57:
```suggestion
ValidMind offers support for both first-party models and [third-party vendor models](../about/glossary/glossary.qmd#vendor-model).
```

nrichers (commented):
Left a couple of suggestions to use relative links.

validbeck (approved):
@nrichers I'm observing the same behaviour as when I last worked with the Glossary, see sc-4381. Otherwise it more or less works as expected!

PR #206, opened 2024-05-29
Quick fix for sc-3324

validbeck:

> sc-3324 was actually covered by us needing to lay the fountain for sc-4186, so I've just made some modifications to  ["Register models in the inventory"](https://docs.validmind.ai/guide/register-models-in-model-inventory.html) as I noticed the `What's next` link leads you to the incorrect WIP page instead of the correct "Edit model inventory fields" page.

#### Onboarding groups
Not sure what happened, but this got some additional text slapped on the top above the title that causes the title to render incorrectly so I've fixed that as well.

| Old | New |
|---|---|
|Screenshot 2024-05-29 at 10 39 08 AM|Screenshot 2024-05-29 at 10 39 22 AM|


validbeck:
> Fixes to QMD files look great, but I'm not sure why changes to notebooks would be included in this PR?

@nrichers I didn't make any changes to any notebooks, weird. 🤔 I did call `make get-source` though?

nrichers:
> I did call `make get-source` though?

That would do it. My suggestion would be to keep PRs as atomic as possible as every change we introduce has the potential to break stuff and needs to be reviewed.

validbeck:
> My suggestion would be to keep PRs as atomic as possible as every change we introduce has the potential to break stuff and needs to be reviewed.

Gotcha, I was under the impression that calling `make get-source` to make sure we have the most recent version of the `developer-framework` files was the SOP. Is it good to go otherwise? Should I just refresh this PR somehow (just make the edits again fresh without retrieving the files from `developer-framework`?

nrichers:
> I was under the impression that calling `make get-source` to make sure we have the most recent version of the `developer-framework` files was the SOP.

We have an SOP for publishing the latest release on our docs site, complete with release notes. If you're just making an interim update, I'd keep that as small as possible. (We should probably have a process doc for interim updates, eh?)

> Should I just refresh this PR somehow (just make the edits again fresh without retrieving the files from developer-framework)?

You can checkout folders from other branches to undo changes which is helpful in this case: `git checkout main -- notebooks`. I pushed https://github.com/validmind/documentation/pull/206/commits/439d0bdc5707afc7f176e7013a9a8a1f9c16c6fd to show how that works. 

(You'll still need to render the site to update the output.)

nrichers (approved):
LGTM after you run `quarto render` and commit the output. 🚀

validbeck:
> LGTM after you run quarto render and commit the output. 

Did I forget to do that? 💀 I feel like I did that, how can you check? (So I know how to catch it for future reference.)

nrichers:
> Did I forget to do that? 💀 I feel like I did that, how can you check?

More like: I undid the changes to the notebook source but I intentionally did not re-render the site as I wanted you to be able to see only the changes introduced by the `git checkout main -- notebooks` command.

validbeck:
> More like: I undid the changes to the notebook source but I intentionally did not re-render the site as I wanted you to be able to see only the changes introduced by the git checkout main -- notebooks command.

Makes sense, does this look about right? 😓 

Screenshot 2024-05-29 at 3 52 18 PM

nrichers:
> does this look about right? 😓

Looks close enough! I'm sure that's fine.

PR #218, opened 2024-06-14
Updated test-driven blocks

noosheenv:

To keep up with the new test-driven block menu, I updated the docs:
- Added a screenshot of the menu that pops up for test-driven blocks
- Updated instructions to match current features

| Before | After |
| ----- | ----- |
| image | image |


site/guide/work-with-content-blocks.qmd
```diff
@@ -41,11 +41,18 @@ Test-driven block
...
-    
-7. For test-driven blocks, a preview of the available metrics or threshold test gets shown. Click **Insert module** when you are ready.
+    - **Test-driven block**: Adds a new section with test results. 
+
+7. For test-driven blocks, a menu containing available tests will appear:
+
+
+      ![](test-driven-block-menu.png){width=85%}
+
+   - Select tests using the checkboxes on the left.
```
nrichers, line 51:
```suggestion
   - Select the tests to insert into the model documentation.
```

A good rule of thumb is to AVOID describing how the UI works but focus on the user task (here selecting tests to insert). Typically, if you have to describe the UI, there's a design issue.

site/guide/work-with-content-blocks.qmd
```diff
@@ -41,11 +41,18 @@ Test-driven block
...
+   - Search by name using **{{<fa magnifying-glass >}} Search** on the top-left.
+   - Click on a test to preview it. By default, selected tests are previewed.
```
nrichers, line 53: This text does something different than the other bullets — maybe make this a follow-on paragraph at the same indentation level as the bulleted list that says something like "To preview what is included in a test, select it. By default ..."

(AFAIK, selecting means clicking, so we should consistently use the same simplified action words.)

site/guide/work-with-content-blocks.qmd
```diff
@@ -41,11 +41,18 @@ Test-driven block
...
+
+8. Click **Insert test result to document** when you are ready.
```
nrichers, line 55:
```suggestion
8. Click **Insert _n_ test result to document** when you are ready.
```

It can be hard to match buttons that change based on what you select but this gets you a bit closer.

nrichers (commented):
Nice work! I left some smaller comments — let me know when these are addressed and I'll approve. 🎉🎉🎉 

One other request: Can you clean up the screenshot so there is no background content included around the edges, please? Screenshots should include only the immediately relevant content and any unrelated bits should be cropped out.

noosheenv:
Thanks for the feedback and tips, Nik! I implemented the changes in the latest commit. Here is a sneak-peek:

image

nrichers (approved):
LGTM (looks GREAT to me)! Thank you for making these changes. Ship it! :shipit:

PR #221, opened 2024-06-19
Improve workflows prerequisite info

nrichers:

This PR makes an incremental improvement to our developer and validator training to be more explicit about workflows needing to be configured, which was a review comment from Michael: 

- In the developer training module, **add a new tab with instructions for verifying the approval workflow**, similar to what the validator training already has. (This fixes an omission in the original training module for developers.)
- In both the developer and validator training modules, **mention that workflows must be set up by an administrator**.
- In the linked-to topic for verifying workflows, **add a prerequisite for workflow setup with a link to the new docs for setting up the workflow in the callout**, now that those docs are available.
- **Replace a very stale image of the default workflow** with an updated version from the current UI

## Output changes 

- In the developer training module, **add a new tab with instructions for verifying the approval workflow**:

    image

- In both the developer and validator training modules, **mention that workflows must be set up by an administrator**:

    image

- In the linked-to topic for verifying workflows, **add a prerequisite for workflow setup with a link to the new docs for setting up the workflow in the callout**:

    image

**Replace a very stale image of the default workflow**:

| Before | After |
| --- | --- |
| image | image |


validbeck (changes requested):
Looks more or less good to me, but I would honestly just remove the screencap under the "Click ... See workflow to visualize the entire workflow that this model will go through." line as Workflows are fully custom, unless you want to tag on a `This workflow is custom, and will be configured by an administrator to reflect your team's processes and may not appear like the example below` or something like that. My preference is just to remove the image and not overcomplicate things (and saves us time if this default starts to change in the future as it likely will etc.)... especially as there already is a default image in the "Learn more" link!

**EDIT:** I see now that you single-sourced the instructions that have that default image. Maybe take the image out of the source and leave it on the main page?

nrichers:
> I see now that you single-sourced the instructions that have that default image. Maybe take the image out of the source and leave it on the main page?

I'll have to test if this can be made to work as expected. The conditional content itself is not the issue, it's when you try to include _conditional content in a numbered list_ — it tends to reset the numbering regardless of what tricks I've tried. Lemme see what I can do; if it's not possible, I'll just remove the screenshot as suggested.

validbeck:
Hm, in my view the image is at the end of the list of instructions so I don't think it will necessarily be an issue but maybe we are looking at different things? 🤔 

| Before | After |
|---|---|
| Screenshot 2024-06-19 at 3 59 03 PM | Screenshot 2024-06-19 at 4 00 07 PM |

nrichers:
> Hm, in my view the image is at the end of the list of instructions so I don't think it will necessarily be an issue but maybe we are looking at different things? 🤔

Yes, in fact I just moved the image out into the main topic rather than having it conditionally excluded in the single-sourced content. Problem avoided.

validbeck (approved):

PR #225, opened 2024-06-21
Reimplement docs site main page in css grid

validbeck:
#### index.qmd landing page
- Uses only our default `styles.css` (which I modified for the custom elements on this page) and a tiny bit of custom css for the special links/buttons on this page and Tachyons!
- Now all of the links are the "good" (`.qmd`) ones, except for the ones in the buttons.
- It's not a complete recreation, but I think it looks pretty clean. However, I don't know what it looks like on bigger resolutions, so please preview it and let me know...

| Old | New |
|---|---|
| old-landing-navbar  |  new-landing-navbar |

- I changed the search's type to `overlay`
- I toggled the `show-item-context` to `true`, which gets us the breadcrumbs on the search for the pages!
- I had to remove the search box from the landing, but I added a call-out (and a keyboard shortcut!) to direct users to the search in the top-right corner now present on the top navigation on any page. That being said, if you can figure out how to hack this (because stealing the HTML from the old page didn't work), please do!

| Old | New |
|---|---|
|  Screenshot 2024-06-21 at 4 48 55 PM |  search |

#### Header 
- I replicated the padding seen in my 1st test PR #164
- It also now performs the same across the site except for the `customize-documentation-templates.qmd` which has the extra CSS on top of it... hm. Maybe another Story to investigate, but it's honestly not noticeable unless you're watching it. 

| Old | New |
|---|---|
|  Screenshot 2024-06-21 at 4 49 49 PM | Screenshot 2024-06-21 at 4 49 41 PM  |

#### Footer
- I replicated the padding seen in my 1st test PR #164
- I removed the "Edit this page" and "Report an issue" as this is sort of superfluous on this landing by toggling the `repo-actions` to false
- Following our own style guide rules about italics here ;)
- Resolved the weirdly coloured text

| Old | New |
|---|---|
| Screenshot 2024-06-21 at 4 50 34 PM |  Screenshot 2024-06-21 at 4 50 51 PM |

### Caveats
- This version isn't exactly responsive, but the older one wasn't too good either. This is just an issue with CSS grid, so it applies to anywhere else we use the grids in the docs site. I think we may need to look into a flex-box solution, but maybe that can be another Story?
- I wonder if we can use non-html buttons so we can use the good links everywhere, but so far I haven't figured it out.


nrichers (changes requested):
Kudos for this PR! I'm glad you're getting some experience with Tachyons CSS and this PR generally does a great job of addressing the underlying issue we're having with our landing page, plus you are fixing or improving several other things. 

One request before I test this PR more thoroughly, related to these changes:

> I changed the search's type to `overlay`
> I had to remove the search box from the landing

Can we put the search box on the landing page back to where it was, please? This also entails staying away from `overlay`, for good reason.

I'll explain:

- **Good, easy search is a big deal**: We should aim for the front page to become more functional over time, not less, and that includes giving people the easiest access to our search possible, with the fewest clicks. I'd argue the searchbox is one of the most important features on this page — I use it myself all the time, for example ("I wonder what our docs say about x?" --> docs.validmind.ai --> type into search box). The workaround of telling people how to search from the front page isn't really a solution because you're replacing functionality with an explanation for how our docs UI works — there's an important clue right there.

   (In contrast, including more context in the search results via this PR is a positive change, very much in the right direction.)

- **Let's avoid adding complexity** (phrased positively: add simplicity): The only reason you've had to remove the search box is because of `overlay`, but there's a usability trap hidden in that option that prioritizes aesthetics over usability. As much as I like the look of `overlay` and tested it with our docs site before, you now need to click before you can search, on every docs page, for every search. Every. single. time. An extra click every time you want to search is a small usability anti-pattern with a big impact over time, and it's completely unnecessary since a better option is readily available and already in use on our docs site.

   (Wherever possible, we should simplify the experience, similar to how getting to our recommended QuickStart now takes two clicks instead of five.)

> I wonder if we can use non-html buttons so we can use the good links everywhere, but so far I haven't figured it out.

Simplified HTML buttons for sure, non-HTML with better linking possibly — I did experiment with what's available during the initial training module development a bit. 

We should be able to use [basic rounded buttons](https://tachyons.io/components/buttons/basic-rounded/index.html) but our Quarto styling gets in the way, as you can see in these examples:

```
<button class="{ white bg-pink br3 bn f2 pv2 ph3 mt5 mb5 }" onclick="window.location.href='https://app.prod.validmind.ai/'; return false;">Log in</button>

:::{.dib .white .bg-pink .br3 .bn .f2 .pv2 .ph3 .mt5 .mb5}
[Log in](https://app.prod.validmind.ai/)
:::
```

Capto_Annotation

The reason Quarto looks different is that it's not really a button as `:::` is the equivalent of a div. I'm inclined to say even the HTML button in this example would be an improvement over the much more complex buttons we use today. Or, we could use a slightly different button in Quarto that works well with our existing CSS styling if we can figure out the vertical text alignment.

nrichers:
Did a bit more experimenting with links, here's how to do a relative link in QMD with better alignment and dimming when you hover: 

```
::::{.dib .f3 .pink .bg-white .ba .bw1 .br3 .b--dark-pink .dim .pt3 .ph3}
[Learn more](../guide/get-started.qmd) 
::::
```

image

Still has an underline when you hover, though.

nrichers (approved):
LGTM! 🚀 I pushed https://github.com/validmind/documentation/pull/225/commits/a9dfc8718ad43417c749f71ebf6cc476a010005b to fix a couple of issues:

- The developer & validator links were broken, fixed (index.qmd is already a level above guide/ so no need for the `../`).
- The images were missing `fig-alt` text, might as well fix while I was reviewing. 

My only other comment is that it feels like there is just a bit too much whitespace in the navbar now. E.g. on a 1440x900 laptop screen, it feels like we're conceding just a bit too much whitespace to the words that matter, to the point where it detracts from the reading experience. Up to you if you want to make any additional changes or leave as-is. though.

image

(Granted, I feel about whitespace in the context of words much the same way some people might feel about the libretto in opera — it's just an excuse for the music. Too little and it's not opera; too much and it gets in the way. :P )

validbeck:
> The developer & validator links were broken, fixed (index.qmd is already a level above guide/ so no need for the ../).

Good catch, thanks!

> The images were missing fig-alt text, might as well fix while I was reviewing.

Yeah, I was gonna rename + tag these at some point after our filename convention agreement, but thanks for doing that in the meantime!

> My only other comment is that it feels like there is just a bit too much whitespace in the navbar now. E.g. on a 1440x900 laptop screen, it feels like we're conceding just a bit too much whitespace to the words that matter, to the point where it detracts from the reading experience. Up to you if you want to make any additional changes or leave as-is, though.

I would rather leave it as is, as I introduced this white-space deliberately with the idea that the old navbar was a bit too cluttered for my poor dyslexic eyes (even now, after a few months in where I ostensibly know where to look and where things live!). There are smarter people than me who can more effectively argue why white-space is important, _especially_ on screens (ever try to read a really squished PDF or one really big rambly paragraph like this one, for that matter?), to direct the flow of information and give users room to differentiate between items, for example. I think the amount it has now says "this is a navbar" whereas before it blended into the rest of the site with little visual distinction which is why I didn't even notice it was there the first few times I perused the docs. 

I will render this PR and merge it after I'm done with the template edits for Barclays today, thank you for your thorough review!

nrichers:
> I would rather leave it as is, as I introduced this white-space deliberately with the idea that the old navbar was a bit too cluttered for my poor dyslexic eyes

Works for me!

PR #230, opened 2024-06-26
Updated Swap Documentation Templates for Barclays

validbeck:

> For sc-5129, I quickly edited and fleshed out `swap-documentation-templates.qmd` as this guide was slightly incomplete.

- The old intro was worded in such a way that implied you could make edits TO the template (rather than just switch the parent template), so hopefully that's been clarified in this version. 

| Old | New |
|---|---|
| Screenshot 2024-06-26 at 11 46 33 AM | Screenshot 2024-06-26 at 11 46 43 AM |
| Screenshot 2024-06-26 at 11 47 10 AM | Screenshot 2024-06-26 at 11 47 17 AM |

| Old | New |
|---|---|
| Screenshot 2024-06-26 at 11 47 56 AM | Screenshot 2024-06-26 at 11 48 03 AM |

| Old | New |
|---|---|
| Screenshot 2024-06-26 at 11 50 04 AM | Screenshot 2024-06-26 at 11 50 20 AM |

### Other comments
- Added a quick "What's next" section as per our convention
> When swapping templates, only the document structure is changed. Any modifications that you might have made to content is preserved inside each content block or section:
>- I'm sure I can stress-test this, but what happens to contents in blocks/sections that are removed from the template? 😬 (I assume they are also removed...)


site/guide/swap-documentation-templates.qmd
```diff
@@ -3,9 +3,10 @@ title: "Swap documentation templates"
 date: last-modified
 ---
 
-Swapping templates allows you to switch to a completely different template, upgrade to a more recent version of your current template, or make changes to both the template and template version at the same time.
+Easily swap between different versions of your model documentation or validation report templates within the {{< var vm_platform >}}. Switch to a completely different template, or apply another version of your current template.
```
nrichers, line 6: Re: "Easily swap": Still working on my review, but I think we should chat about our use of words like "easily", "convenient", or "simply" as they are usually removed by technical editors I have worked with. The problem is that these words are vague and sit somewhere outside of the realm of technical communication — what's easy to you might not be easy to me and they aren't measurable. What's worse, if a user tries something, can't perform a task, and then sees our claim that it's easy, it _really_ doesn't help. Best to avoid, I think?

validbeck: That makes sense — you'll have to forgive me, I spent the last half-decade in martech land and am still shaking some of it off 😆

site/guide/swap-documentation-templates.qmd
```diff
@@ -64,31 +65,38 @@ When swapping templates, only the document structure is changed. Any modificatio
...
 
 9. Enter a note to enable completing the swap and click **Swap Template**.
 
-After your model documentation template has been swapped successfully, you can continue to work on your model documentation
+::: {.column-margin}
+^2^ [Customize documentation templates](customize-documentation-templates.qmd)
+:::
 
-#### Reusing old content
+![](swap-template-version.gif){fig-alt="A gif showcasing swapping a documentation template version."}
```
nrichers, line 83: I know we already have the workflows docs where the animated GIF appears in a similar location after the steps, but it feels like an animated GIF that covers ALL THE STEPS described in this section should perhaps live BEFORE the steps? That's where we'd place a 'show me how' video if we had one, I think. 

Maybe something to chat about by way of establishing a convention.

nrichers (approved):
LGTM! Just some minor comments about verbiage and animated GIF placement.

site/guide/swap-documentation-templates.qmd
validbeck: Hm, good point. I sort of just placed these based on some existing docs (that I didn't write). But my personal preference is after, but I'm not a videos person is maybe why — I tend to want to read the instructions myself first, then look for clarification via different methods after if the wording is unclear. Generally screencaps do go AFTER the written word though. 🤔 (And it doesn't always capture the ENTIRE flow, usually just the END of the flow in the case of these templates, the bit with the big menu that comes up that might be confusing.)

PR #240, opened 2024-07-04
Added `scroll-padding-top` to offset anchors for fixed nav

validbeck:

> For sc-4381, I adjusted the behaviour of our anchor links to accommodate for the fixed top navigation. There's so much room because the mobile version of the top-nav is deeper than the desktop version.

| Old | New |
|---|---|
|old-anchor-spans|new-anchor-spans|

### Span highlighting
I adjusted the `<span>` anchors I found to have this neat highlighting that draws attention to the active span!

span-highlighting


site/styles.css
```diff
@@ -13,6 +13,16 @@
   -webkit-font-smoothing: auto;
 }*/
 
+html {
+  scroll-padding-top: 120px;
+  scroll-behavior: smooth;
+}
+
+span:target {
```
nrichers, line 21: Can we remove this anchor styling? It's not meaningful for a _link target_ to be highlighted.

nrichers (changes requested):
This PR does a nice job of fixing the anchor offset issue we were having. My only request would be that we not add highlighting around link targets, I don't think it adds anything that a user would need to know.

site/styles.css
validbeck: My logic here is without it sometimes it doesn't actually jump to the actual anchor exactly — since the offset is not perfect. But at least this way it doesn't _hide_ the anchor with the navbar. I've removed the cool pink styling, though. :((((

nrichers (approved):
My apologies for asking you to remove the cool pink colours from the PR. (Just commented them out, eh? 😁 )

Boring people say LGTM! :shipit:

PR #242, opened 2024-07-05
Upgraded margin notes options — please vote!

validbeck:

> For sc-5288, I looked into some options for cleaning up our margin notes experience for both authors and users.

### Options

Here are the options I whittled it down to, so please vote on your favourite before I make the changes globally across the site:

| Hover only | Margin only | Margin & hover |
|---|---|---|
|Screenshot 2024-07-05 at 10 27 42 AM | Screenshot 2024-07-05 at 10 28 13 AM| Screenshot 2024-07-05 at 10 28 31 AM|

Screenshot 2024-07-05 at 10 37 55 AM


noosheenv:
This is so cool, Beck! I like them all but the vibes for hover-only is my favourite. Also, the testing page thing is very cool and we should totally demo that on Monday 👀

nrichers (commented):
Based on the source I see, there is **no difference in authoring cost**, correct? I.e. this is strictly a question about how we style these links. 

Also, I have a question about the examples: why use numbering like `[^2]`, followed by a separate footnote? AFAIK `^[FOOTNOTE-TEXT]` works just fine, at least in revealjs presentations where I have used that, and it takes the manual footnote numbering work out of the equation. 

Oh, and to answer your question: I genuinely like how references now appear in the margins — it's a great look for our docs that's different while also feeling very familiar and mature overall. My vote would be for **margin-only** or **margin & hover**.

validbeck:
> Based on the source I see, there is no difference in authoring cost, correct? I.e. this is strictly a question about how we style these links.

It's the same amount of work for me, yes. 

> Also, I have a question about the examples: why use numbering like [^2], followed by a separate footnote? AFAIK ^[FOOTNOTE-TEXT] works just fine, at least in revealjs presentations where I have used that, and it takes the manual footnote numbering work out of the equation.

I'll try that again, but when I tried that it didn't really work when there were multiple footnotes on the page.

PR #249, opened 2024-07-15
Update site banner

nrichers:

Love our new banner! Here's some small text tweaks I'd like to suggest: 

- Group like things together to reduce segmentation ('gain hands-on experience & training env' go closely together)
- End the text cleanly with the main CTA ('explore what VM has to offer')
- Remove period at the end as banners typically shouldn't have one, similar to UI toasts 

EDIT: An even simpler banner would be "**Try our training environment to explore what ValidMind has to offer**" — what do you think?

EDIT EDIT: I'm gonna push a commit for the even simpler banner.

### DIFF
--- a/site/_quarto.yml
+++ b/site/_quarto.yml
@@ -3,7 +3,7 @@ project:
   post-render: make python-docs
 website:
   announcement: 
-    content: "[**{{< fa graduation-cap >}} ValidMind Academy**](/training/training-overview.qmd) — Gain hands-on experience with our training environment and explore what ValidMind has to offer" 
+    content: "[**{{< fa graduation-cap >}} ValidMind Academy**](/training/training-overview.qmd) — Try our training environment to explore what ValidMind has to offer" 
     position: below-navbar 
   favicon: validmind.png
   title: "ValidMind"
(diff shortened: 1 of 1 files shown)


validbeck (approved):
> Try our training environment to explore what ValidMind has to offer

I like this! Either works for me, I just stole the first line from the training landing page when I made the announcement 😅 

> Remove period at the end as banners typically shouldn't have one, similar to UI toasts

Is there some documentation on this (or some style guides that follow this convention) so I can read up on it to familarise myself? It just looks strange without the period here but I'm sure you're correct!

nrichers:
> Is there some documentation on this (or some style guides that follow this convention) so I can read up on it to familarise myself? It just looks strange without the period here but I'm sure you're correct!

When I spun up the UX team at Elastic, I got my lead UX writer to work on related writing guidelines for the designers: https://eui.elastic.co/#/guidelines/writing/guidelines#punctuation — look under "Don't use unneccessary punctuation".

nrichers:
> **EDIT:** For some reason the `callout-appearance` displays a bit strangely with the CSS when it's not set to `simple`, BTW. 🤔 It was intermittent when I tried it, though, like it happened once then not again.

I didn't touch this, looks like GitHub just stripped out an extra line break?

PR #251, opened 2024-07-16
Add Private Service Connect & rework login info

nrichers:
## Internal Notes for Reviewers

Highlights of this PR:

**Private Service Connect**

- Added topic for configuring Google Cloud Private Service Connect:
   
   image

**_variables.yml**

To reflect that there are now multiple production login URLs: 

- Split `vm_url` into `vm_url_us1` and `vm_url_ca1`.

   | Before | After |
   |--------|-------|
   |  `vm_url`: "https://app.prod.validmind.ai"     |    `vm_url_us1`: "https://app.prod.validmind.ai"<br>`vm_url_ca1`: "https://app.ca1.validmind.ai"   |

- Removed `vm_login` variable as no longer applicable in a multi-production sites world

**Step 1 in task topics**

To enable a log in step at the beginning of task topics that works for multiple production URLs:

- Re-added login topic, now updated with info about public and private URL info about accessing the ValidMind Platform UI

   image

- Replaced step 1 in task topics with login link to point to login topic instead of platform UI URL

   | Before | After |
   |--------|-------|
   | image | image |

**CTAs, training, and other links** 

- Updated signup CTAs to use new `vm_url_us1` variable (it's the same URL we've always sent people to)
- Updated training overview to use `vm_url_us1` (training modules are unchanged and use HTML links that point to the same URL)
- Updated release notes links to use `vm_url_us1` as the default

**Stale content**

In an effort not to maintain content that is defunct: 

- Deleted the old `guide/get-started-sandbox.qmd` file as it's been superseded by our ValidMind Academy content
- Deleted the defunct `site-unused/qmd/get-started-closed-beta.qmd` file as we're done with the beta now

**Open items** 

- [x] As an experiment, is there a step 1 option that  feels more elegant? **>>> TO BE MOVED INTO A SEPARATE PR**

   **With [Nutshell](https://ncase.me/nutshell/)**, via a single-sourced step-1.qmd:

   

   **Notes**
   - This single-sourced step cannot start with `_` or the extension does not work, meaning there will be a separate step-1.html output topic
   - The Nutshell popups are SLOW 
   - The Nutshell info icon text in the top right feels too playful for an enterprise product, might need to hide
   - Some additional styling would be needed for this to look right

   Overall, I'm not sure it's an improvement over linking to the login topic. 

   **With `citationshover: true`** we could do something like this: 

   image

   **Notes**
   - Very fast, clean look
   - Enabled by default, no extension needed 
[...]

PR #256, opened 2024-07-18
Add basic Nutshell example for experimentation

nrichers:

This PR does exactly one thing: it provides a working Nutshell example that can be used to experiment with a better alternative for step 1 in task topics. @noosheenv @validbeck 

Includes:

- A `step-1.qmd` file with text for the Nutshell bubble (dependency on #251 to render variables, FYI)
- A modified `guide/configuration/managing-your-organization.qmd` task topic that demonstrates how to use the Nutshell bubble
- Some initial style modifications in `styles.css` (hide the text with the source, hide the icon, pink links, smaller bubble radius) — more is needed

Note that tables don't seem to render correctly.

### DIFF
--- a/site/_log-in.qmd
+++ b/site/_log-in.qmd
@@ -1,0 +1,2 @@
+##### {{< fa circle-exclamation >}} Before you begin
+Make sure you are [logged in to the ValidMind Platform UI](/link-to-login-page.qmd).
--- a/site/guide/configuration/managing-your-organization.qmd
+++ b/site/guide/configuration/managing-your-organization.qmd
@@ -11,6 +11,8 @@ listing:
     contents:
     - set-up-your-organization.qmd
     - configure-aws-privatelink.qmd
+filters:
+   - nutshell
 ---
 
 Access to the {{< var validmind_platform >}} is associated with an organization, which encompasses all your users, groups, and business units. As a user, you can belong to multiple organizations.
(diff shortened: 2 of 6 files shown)


validbeck:
Some initial thoughts:

- If we do end up using the Nutshell extension, I'd recommend a top-level folder in `site` to the effect of `nutshell` and then naming the file/instruction with the contents instead of a floating unclear `step-1.qmd` etc. 
- We might also want to change the whole instruction to a variable e.g.`[:Log in to ValidMind](/nutshell/login.qmd#x-step-1).`
- I was able to get the table to perform a bit better with some non-breaking spaces, but the solution isn't ideal...
- The little dots animation prefacing the "link" seems to be baked into the JS file, and I'm sure we could modify it (perhaps to 3 lines of defense, converging into a checkmark on hover...?) but it does seem sort of juvenile overall in contrast to the rest of the site. 

Screenshot 2024-07-30 at 10 04 41 AM

> Overall, after playing with it, and on consideration of "less links in the instructions and simpler is better for both author and user," I think I'd rather us resurrect a `Log in to ValidMind` guide that is linked in the **Prerequisites** section of each relevant page. 

- For example, we have a "You are logged into the ValidMind Platform UI" list item already in some articles such as `Collaborate with others`. 
- This would greatly cut down on the loginloginlogin links on pages with many sets of instructions, and allow us to update one central page with less fuss re: login instructions.

nrichers:
Closing — while we liked what Nutshell could do, it's not quite what we are looking for. This work will feed into a redesigned "Prerequisites" section as proposed by @validbeck instead.

PR #263, opened 2024-08-01
Updated variable naming conventions

validbeck:

> For sc-5612, I decided to ask for forgiveness rather than permission 😆 

new-rendered-variables

See new file: https://github.com/validmind/documentation/blob/beck/sc-5612/update-variable-naming-conventions/site/_variables.yml

#### Log into ValidMind
- I changed this to `Log in to ValidMind`, as the convention around log in (the verb) and login (the noun) is always "log in," and not "log into."
- I also added a variable back in for this to link to this corrected page on instructions.
- I moved this OUT of "Managing your organization," as this is actually the first step to getting anywhere in the platform. 
- There is now also a "What's next" section, directing you to organization setup, user management, and profile viewing

| Old | New |
|---|---|
| old-log-in | new-log-in |

#### Managing your organization
- Updated the"What's next" section to include the new Google VPC page

| Old | New |
|---|---|
| old-managing-org | Screenshot 2024-08-01 at 2 20 54 PM |

**EDIT:** Added the login page to the Guides landing:

| Old | New |
|---|---|
| Screenshot 2024-08-06 at 10 04 44 AM| Screenshot 2024-08-06 at 10 04 07 AM|


site/_variables.yml
```diff
@@ -1,44 +1,74 @@
...
+# Amazon Web Services Privatelink specific URLs
+# Rendered format: `{{< var aws.privatelink >}}`
+
+aws:
+  privatelink: "https://private.prod.vm.validmind.ai"
+
+# Google Cloud Private Service Connect specific URLs
+# Rendered format: `{{< var psc.validmind >}}`
+
+psc:
```
nrichers, line 55: This `psc` acronym doesn't quite make sense to me and I don't think anyone will ever remember it. The convention you follow on lines 49-50 feels more elegant as it starts with the cloud provider and then works its way down to the service. 

How about this? 

```yml
gcp:
  private_service_connect: "https://private.us3-prod.validmind.ai"
  private_authentication: "https://auth.private.validmind.ai"
```

(FYI, Spencer has said that the authentication endpoint might go away at some point in the future.)

nrichers (commented):
Is it weird that I was reading through the Quarto docs over the holidays and ran across the same page for variable arrays? They remind me of the [compound or stemmed variables](https://www.ibm.com/docs/en/cics-ts/6.x?topic=data-using-compound-variables-stems) I used when I worked on IBM z/OS mainframes many, many years ago. 

This PR is a great thing, my only comment is around the variables for Google's Private Service Connect. Not sure if you want to discuss further?

site/_variables.yml
validbeck: I was torn between `psc` and `gcp` or `vpc`... the last one seems more generic, but then I feel like the aws should go under that as well. What about:

```
gcp:
    connect: ...
    authentication: ...
```

I'm really not a fan of the super long snake_case ones, as that feels even MORE tedious than just typing out the values in some cases! I had a version where I had a "rule" where I discussed snake_case, but the more I thought about it the more it went against us trying to simplify it...

nrichers: > What about:
> 
> ```
> gcp:
>     connect: ...
>     authentication: ...
> ```
> 
> I'm really not a fan of the super long snake_case ones, as that feels even MORE tedious than just typing out the values in some cases! 

+1 to this suggestion — both the cloud provider context and the fact that it's a private connection are clear. Agreed on avoiding long variable names.

nrichers (approved):
Latest changes look great. Ship it! 🚢

nrichers:
> I changed this to Log in to ValidMind, as the convention around log in (the verb) and login (the noun) is always "log in," and not "log into."

One postscript on this, though the upshot is that I don't actually mind either way: "Log in" is the the verb form, true, but both "log in to" and "log into" are valid if slightly different grammatical constructions. In a previous role, we had a huge discussion about the difference among the writers — I tend to err on the side of simplicity and typically have used "log into" but [this comment](https://english.stackexchange.com/questions/5302/log-in-to-or-log-into-or-login-to#comment1188804_5303) is a good summary of why either works.

PR #267, opened 2024-08-08
Updated Jupyter Notebooks to match prod branch

validbeck:

> [Remember our very tired day where we kept needing to adjust the new names of notebooks](https://validmind.slack.com/archives/C04U7TKHH62/p1721258056890579?thread_ts=1721257654.158869&cid=C04U7TKHH62)?

We updated the notebooks for the `prod` branch (so the live site is fine) but our current `main` branch is now behind on the notebook renames so any local previews / renders won't show the corrected notebooks. I've fixed this by pulling in the latest notebooks from `developer-framework`.


nrichers (approved):
LGTM — also checked that the path name for NLP and LLM notebooks in the `main` branch is correct (it is). 🚀

PR #269, opened 2024-08-09
Fix `nlp and llm` path in output 

nrichers:

When I reviewed https://github.com/validmind/documentation/pull/267, I checked the `site/` source path was correct. I did not check the `site/_site` output path and, lo and behold, it's still cased incorrectly on `main`. This PR fixes that issue, hopefully once and for all.


nrichers:
> This pull request removes several generated HTML files from the repository. These files are located in the `site/_site/notebooks/code_samples/nlp_and_llm/` directory and include various demo files related to NLP and LLM integrations and summarizations. The removal of these files helps to keep the repository clean and reduces unnecessary clutter from generated content.

Interesting that the bot gets this quite wrong, given that it's clearly a rename and not a delete. 

image

validbeck (approved):
Looks good to me — did you just directly rename the output folder?

nrichers:
> Looks good to me — did you just directly rename the output folder?

Two-step rename, as before. Seems safest.

PR #326, opened 2024-09-04
Delete site/_site directory in staging

nrichers:

This PR removes the `site/_site` folder from the `staging` branch. This folder is no longer be used as we now render the docs site in the context where output is needed.


validbeck (approved):
Looks good! Let's gooo

PR #309, opened 2024-09-03
Add staging docs site deploy and update prod deploy

nrichers:
## Internal Notes for Reviewers

This PR includes:

- `deploy-docs-staging.yaml` — a new GitHub workflow to render & deploy our staging docs site whenever a commit to the `staging` branch is made
- `make deploy-staging` — a new Makefile action to do the same locally
- `deploy-docs-prod.yaml` — an updated GitHub workflow for `prod` to render the site before deploying

In our new docs CI/CD deployment infra, this PR implements the following highlight: 

image

### `deploy-docs-staging.yaml` 

Successful run: https://github.com/validmind/documentation/actions/runs/10704868760

**Note**: The `staging` S3 bucket is not yet available via DNS, so the way to verify this workflow works is to look at the output of the workflow run. 

Here's a screenshot that shows the rendering of the docs site passed and the beginning of the deploy via `aws s3 sync`:

image

And here's a sceenshot that shows the end and the clearing of the cache with `cloudfront create-invalidation`:

image

### `make deploy-staging`

`make deploy-staging` abridged output when run locally:

```
❯ make deploy-staging

Deploying staging site ...
[  1/313] guide/configuration/set-up-your-organization.qmd
[  2/313] guide/configuration/configure-google-private-service-connect.qmd
[  3/313] guide/configuration/manage-roles.qmd
...
[311/313] get-started/platform/explore-sample-model-documentation.qmd
[312/313] get-started/platform/next-steps.qmd
[313/313] get-started/quickstart.qmd

make

Updating Python documentation ...

Output created: _site/index.html

upload: _site/404.html to s3://docs-ci-cd-staging/site/404.html                        
upload: _site/about/contributing/join-community.html to s3://docs-ci-cd-staging/site/about/contributing/join-community.html
upload: _site/about/contributing/validmind-community.html to s3://docs-ci-cd-staging/site/about/contributing/validmind-community.html
...
upload: _site/validmind/validmind/vm_models.html to s3://docs-ci-cd-staging/site/validmind/validmind/vm_models.html
upload: _site/validmind/validmind/unit_metrics.html to s3://docs-ci-cd-staging/site/validmind/validmind/unit_metrics.html
upload: _site/validmind/validmind/tests/prompt_validation/Robustness.html to s3://docs-ci-cd-staging/site/validmind/validmind/tests/prompt_validation/Robustness.html
```

### `deploy-docs-prod.yaml`

Successful run: https://github.com/validmind/documentation/actions/runs/10705064385

[...]

PR #328, opened 2024-09-04
Delete site/_site directory in prod

nrichers:

This PR removes the `site/_site` folder from the `prod` branch. This folder  is no longer be used as we now render the docs site in the context where output is needed. 

(Following the merge of #309, this folder is ignored, as we render the site first and then deploy the output from the workflow rather than what's been committed.)


validbeck (approved):
🤞🏻

PR #330, opened 2024-09-05
Only running the PR previews bot when a PR is ready for review

validbeck:

> @cachafla The PR Review bot is neat but runs way too often for us. I've adjusted it (hopefully) to only run the once when a PR has a review requested based on these docs: [GitHub — Events that trigger workflows (pull_request)]( https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request)

cc @nrichers for visibility.

### DIFF
--- a/.github/workflows/ai_explain.yaml
+++ b/.github/workflows/ai_explain.yaml
@@ -6,7 +6,7 @@ permissions:
 
 on:
   pull_request:
-    types: [ready_for_review]
+    types: [review_requested]
 
 jobs:
   explain-pr:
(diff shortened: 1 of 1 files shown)


.github/workflows/ai_explain.yaml
```diff
@@ -6,7 +6,7 @@ permissions:
 
 on:
   pull_request:
-    types: [opened, synchronize]
+    types: [ready_for_review]
```
nrichers: I would include at least some additional options here here, as your workflow will run only when the PR goes from draft mode into ready for review.

nrichers (approved):
LGTM! 🚀

cachafla (approved):
Thanks! I also have a couple of improvements planned and hopefully a doc-focused version that should work better for this repo 👌

PR #336, opened 2024-09-06
Modify make python-docs to support rendering online

nrichers:

This PR modifies `make python-docs` to ensure we always make Python docs available for inclusion with the docs site.

**The problem**

Our Python docs are currently missing: https://docs.validmind.ai/validmind/validmind.html (404).

**The solution**

To solve this issue, Python docs are now stored in a new `site/python-docs/` folder, analogous to what we already do for notebooks and test descriptions. The post-render step now copies those files into the docs site output every time. 

Previously, the files had to be available as a result of running `make get-source` (which calls `python-docs`) or had to be already available in  previously committed `site/_site` output as a fallback. If neither of these conditions could be met, the files would be missing.

## How to test

1. Check the preview for this PR — the Python docs are there. 
2. Check the preview for any other PR — the Python docs are missing. 

Ergo, the modified `make python-docs` step that gets run whenever the site is rendered works.

### DIFF
--- a/site/Makefile
+++ b/site/Makefile
@@ -1,6 +1,7 @@
 # Define source and destination directories
 SRC_DIR := _source/developer-framework
 DEST_DIR_NB := notebooks
+SRC_DIR_PYTHON := python-docs
 DEST_DIR_PYTHON := _site/validmind
 DEST_DIR_TESTS := tests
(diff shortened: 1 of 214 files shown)


validbeck (approved):
Woot, I see the folder locally when I pull down the branch as well as the files present in the PR preview site!

Thanks for the quick fix, Nik!

PR #339, opened 2024-09-06
Fix Heap embeds for docs sites — `main` branch

nrichers:
## Internal Notes for Reviewers

This PR is the first out of a series of three pull requests and:

- **Re-adds the missing development tracking file for Heap** — will use this file in the `staging` branch later <br>(Also see #340)
- **Removes production tracking for Heap from `main` (and `staging`)** — we will re-enable tracking for `prod` only later<br>(Also see #342)

PR #340, opened 2024-09-06
Fix Heap embeds for docs sites — `staging` branch

nrichers:

This PR follows #339 and enables tracking for Heap in `staging` using the correct development project in Heap.


validbeck (commented):
How do we ensure that this isn't overwritten automatically when `main` is merged into `staging`?

nrichers:
> How do we ensure that this isn't overwritten automatically when `main` is merged into `staging`?

For the three PRs that I have on the go, you want to merge them in the right sequence: `main`, then `staging`, and then  `prod`. 

After that, the only way this would get overwritten is if someone deliberately re-adds Heap tracking to `_quartyo.yml` on `main`.

validbeck:
> After that, the only way this would get overwritten is if someone deliberately re-adds Heap tracking to _quarto.yml on main.

I guess my confusion is won't what's in `staging` override what's on `prod`? 🤔 I think my hamster brain hasn't made the connections yet.

nrichers:
> I guess my confusion is won't what's in `staging` override what's on `prod`? 🤔 I think my hamster brain hasn't made the connections yet.

Files differing between branches is not a concern, they should be transparent to GitHub unless there is an actual commit that needs to be merged from one branch into another.

validbeck:
> Files differing between branches is not a concern, they should be transparent to GitHub unless there is an actual commit that needs to be merged from one branch into another.

That's what I mean, when we merge from `staging` into `prod` doesn't that unify them?

PR #341, opened 2024-09-06
Fix heap embeds for docs sites — prod branch

nrichers:

This PR follows #339 and #340 and enables tracking for Heap using our production project. (It also adds the missing development Heap file for completeness but it is not used.)


nrichers:
Closing as not actually needed (the missing development Heap file will flow here from staging).

PR #344, opened 2024-09-09
Delete old .github/workflows/render-site.yaml workflow

nrichers:

This PR deletes our old `Run quarto render` GitHub workflow that has been superseded by `Validate docs site` (`validate-docs-site.yaml`), which runs the same workflow but also deploys a PR preview site. This workflow no longer exists in `main` but is still present in the `staging` and `prod` branches. 

As you can see in this PR example, the old check still runs:

image


validbeck (approved):
🚀

PR #334, opened 2024-09-06
Investigate listing page issue 

nrichers:
## Internal Notes for Reviewers

This PR modifies our test description wildcard embeds from `**` to `*.md` to resolve duplicated but blank tiles.

| Before — [docs.validmind.ai](https://docs.validmind.ai/developer/model-testing/test-descriptions.html) | After —  [Preview URL](https://docs-demo.vm.validmind.ai/pr_previews/nrichers/sc-6331/investigate-listing-page-issue-for-test-descriptions/developer/model-testing/test-descriptions.html) |
|--------|--------|
| image | image |

Note this change seems to work just fine for subdirectories.

## Troubleshooting approach

- [x] Turn off echoing the `quarto render` output into a logfile so that we can see the output
- [x] Debugging
   - [x] Test rendering on a local Linux container — output looks fine
   - [x] Run `quarto check` before render — looks OK 
- [x] Add any missing extensions
   - [x] TinyTex — no difference
- [x] Experiment with the source
   - [x] Try removing tabset as the problem occurs only in tabs — no difference
   - [x] Try a directory embed without wildcard — does not work despite what the Quarto docs say
   - [x] Try a `*.md` wildcard embed — works, including for subdirectories

During testing on a local Linux container, the site rendered just fine but if we continue to run into issues that point towards differences with Linux runners on GitHub, we could also try testing on `macos-latest`. For more info, see [Standard GitHub-hosted runners for public repositories](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories).

PR #329, opened 2024-09-05
Release notes for Sept 9 2024

validbeck:
## Internal Notes for Reviewers

> For sc-6089, I gathered two `frontend` release tags that have not yet been externalised as there were no newer ones from `developer-framework` (and thus `documentation`).

**LIVE LINK:** [September 9, 2024](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-6089/release-notes-to-sprint-57/releases/2024-sep-09/release-notes.html)

@robinzimmermann You can just click on the link above to review (as long as @nrichers doesn't overwrite it with his testing), no need to pull the branch down locally if you don't want to.

**Keep in mind that the docs buttons link to the "main" branch of the `docs-demo` site (as they require external links) so they won't match up with the PR preview pages.**

### Reset your model workflow
I added this section at the bottom of the "Working with model workflows" so that there was documentation to link to.

**LIVE LINK:** [Working with model workflows](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-6089/release-notes-to-sprint-57/guide/model-workflows/working-with-model-workflows.html#reset-your-model-workflow)

PR #347, opened 2024-09-09
Pushing changes from `staging` into `prod` 

validbeck:

Includes PRs:

- #309 
- #326 
- #329 
- #330 
- #336 
- #344 
- #334 
- #339 

@nrichers I'm a little confused at this step, some of these changes GH is showing we've already pushed to `prod` before. 🤔 Am I doing it wrong?


nrichers:
> @nrichers I'm a little confused at this step, some of these changes GH is showing we've already pushed to `prod` before. 🤔 Am I doing it wrong?

This might be explained by branch divergence, as we are merging branches with histories that will have some differences. The way I understand this, this state is not a major concern as it does not affect the  outcome of the merge. That said, we can monitor future PRs to see if the list of commits continues to grow.

nrichers (approved):
LGTM. Note merging into `prod` does not generally require approval though it's probably a good idea to scrutinize PRs initially until we become more comfortable with the process.

PR #350, opened 2024-09-10
Pulling in latest changes from PR184

validbeck:
## Internal Notes for Reviewers

> Pulling in latest changes from https://github.com/validmind/developer-framework/pull/184 as an example for our internal Community processes. 

### Retrieve latest from `developer-framework`
Process doc here: https://www.notion.so/validmind/Retrieve-latest-from-developer-framework-c44be5deaf954b99b643b17797b59d89?pvs=4

PR #355, opened 2024-09-12
Delete finding docs update

validbeck:

> For sc-5947, I added a new section at the bottom of "Add or update model findings": [**Delete model findings**](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-5947/documentation-delete-finding-button/guide/model-validation/add-up) (CLICK FOR LIVE PREVIEW)

### Python docs
For some reason GH is registering these files as changed — not sure if that's purposeful, I know we had to have that weird fix in place. I created this branch off of `main`, so it shouldn't be different, but yet... (Maybe it's picking up on the files I generate when I preview? I can reset, but just in case!)


nrichers (changes requested):
The changes generally LGTM, but the topic title and file name are now confusing which I discovered when searching for these changes in the PR preview. We say the topic is about adding or updating findings but, oh, there is now also delete?

image

The easiest might be to 

- Rework the topic title and intro to include deleting
- Update the file name and links to match
  
> For some reason GH is registering these files as changed — not sure if that's purposeful, I know we had to have that weird fix in place.

There shouldn't be any fixes in place that I would call "weird" right now but we do commit the Python docs like anything else we pull out of `developer-framework`. If you created your working branch off `main` without having run `make get-docs` (or having inherited these docs via some other branch you worked on), we should investigate why this is happening.

validbeck:
> We say the topic is about adding or updating findings but, oh, there is now also delete?

To me "delete" is an update you make. I considered putting it under the sub-heading but it didn't seem quite right at the time but if you prefer I can just slot it under there. 

> There shouldn't be any fixes in place that I would call "weird" right now but we do commit the Python docs like anything else we pull out of developer-framework. If you created your working branch off main without having run make get-docs (or having inherited these docs via some other branch you worked on), we should investigate why this is happening.

Correct, I did not call `make get-source`.

All I did for this PR was:

- On main, `git put origin main`
- Checkout to new PR branch
- Edit the one add/update file
- Call `git status` — Python docs come up as changed.

validbeck:
@nrichers Alternate version for your consideration:

Screenshot 2024-09-12 at 2 05 38 PM

nrichers (approved):
"Add and manage" works, thank you for making that update. LGTM. 🎸 

> To me "delete" is an update you make.

CRUD would tell us otherwise: "In computer programming, create, read, update, and delete (CRUD) are the four basic operations of persistent storage." 

They are discrete operations.

validbeck:
> CRUD would tell us otherwise: "In computer programming, create, read, update, and delete (CRUD) are the four basic operations of persistent storage."

Fair enough!

validbeck:
@nrichers Do you want me to reset the `python-docs` to `main` before merging? Still not sure why they were flagged as changed.

nrichers:
> @nrichers Do you want me to reset the `python-docs` to `main` before merging? Still not sure why they were flagged as changed.

I would go ahead and merge. I'm more curious about why these showed up in the first place but nothing you described sounds out of the ordinary. I suspect there was something on your main branch already when you created your working branch but this side of a time machine, it's unlikely we can pinpoint the issue.

validbeck:
Hm, after merging #350 in resetting to `main` gets no changes, and I can no longer replicate the issue on new test branches so I will consider it resolved.

PR #358, opened 2024-09-12
Moved activity to model inventory & adjusted pages

validbeck:
## Internal Notes for Reviewers

> For sc-6501 I updated the Model Activity comment filter to reflect the better brand new drop-down. I also realised that the page was MODEL activity, and not documentation activity, so I renamed it and moved it under "Model Inventory."

Screenshot 2024-09-12 at 2 44 50 PM Screenshot 2024-09-12 at 2 59 46 PM

- I've updated any references to this page & its single-source to the new location/name
- I've added this page to the Guides listing under "Model Inventory"
- I've added the link to this moved file to all the pages under "Working with model documentation" in the `What's next` section, and some other places like model validation & model findings... i.e. anywhere it seemed relevant!

Screenshot 2024-09-12 at 2 59 39 PM

PR #352, opened 2024-09-11
Revamped model findings docs for attachments

validbeck:
## Internal Notes for Reviewers

> For sc-5951, I revamped the old super long "Work with model findings" doc into three separate guides:
> - **Working with model findings:** New overview page
> - **View and filter model findings**
> - **Add and update model findings**

### Working with model findings
- Technically a net-new page, replacing the old "Work with model findings".
- Now includes a **Key concepts** section that needs an accuracy check, since I got a little help from ChatGPT.

> [**LIVE PREVIEW**](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-5951/documentation-implement-add-attachment-to/guide/model-validation/working-with-model-findings.html)

### View and filter model findings

| Old | New |
|---|---|
| View model findings didn't go granular into all findings & model-specific findings. | Added granularity of all findings & model-specific findings to view findings, as well as some example screenshots showcasing the differences between the two views. |
| [OLD SECTION](https://docs.validmind.ai/guide/model-validation/work-with-model-findings.html#view-model-findings) | [LIVE PREVIEW](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-5951/documentation-implement-add-attachment-to/guide/model-validation/view-filter-model-findings.html) |

### Add and update model findings

| Old | New |
|---|---|
| ... | New "Manage supporting documentation" section, detailing how to work with attachments after a finding is added. Added gifs for the sections that were not straightforward. |
| [OLD SECTION](https://docs.validmind.ai/guide/model-validation/work-with-model-findings.html#add-project-findings) | [LIVE PREVIEW](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-5951/documentation-implement-add-attachment-to/guide/model-validation/add-update-model-findings.html) |

### Other cleanup
I noticed a bunch more references to documentation projects etc. that I missed in the first round, so I cleaned all of those up and made sure there weren't any broken links in the shuffle.

PR #351, opened 2024-09-10
Doc: Archive and delete inventory models

validbeck:
## Internal Notes for Reviewers

> For sc-4690, I created a net-new doc: Archive and delete inventory models

### Archive and delete inventory models

> [**Live Preview**]( https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-4690/documentation-archive-delete-models-from/guide/model-inventory/archive-delete-models.html)

### Managing the model inventory 
The new delete/archive page has been added to this landing: [**Live Preview**](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-4690/documentation-archive-delete-models-from/guide/model-inventory/managing-model-inventory.html)

## External Release Notes

You can now archive models when they are decommissioned or delete models that you no longer require records of, keeping your ValidMind inventory accurate and up to date with your organization's current resources.

PR #362, opened 2024-09-13
Publishing changes from PR #340 onwards

validbeck:

- https://github.com/validmind/documentation/pull/350
- https://github.com/validmind/documentation/pull/351
- https://github.com/validmind/documentation/pull/352
- https://github.com/validmind/documentation/pull/355
- https://github.com/validmind/documentation/pull/358


validbeck:
@nrichers Any idea why there's some conflicts?

The `.qmd` files are likely because I moved/renamed those but the `python-docs` output is weird. 🤔

nrichers:
> The `.qmd` files are likely because I moved/renamed those but the `python-docs` output is weird. 🤔

Probably not so much weird as somewhat expected for HTML files. It's a similar issue to what we were having with the HTML output of our docs site before.

My recommendation would be to check out your PR locally, merge in `staging` on the command line with the option `--strategy-option theirs` and just let `git` do its thing. 

Alternatively, you could manually resolve the conflicts yourself but it would also need to be done locally. Happy to help.

PR #368, opened 2024-09-16
Update workflows to trigger on workflow_run

nrichers:

This PR adds a trigger to workflows that deploy our docs sites to staging and prod. Specifically, these workflows now also trigger when the workflows to merge into these branches have completed. This change is required to 

(Note the workflow to merge staging into prod doesn't exist, yet. This is for future me who would otherwise be puzzled why deployment doesn't trigger when we try to automate. If you trace it back to this PR — you're welcome.)


validbeck (approved):
🤞🏻

PR #410, opened 2024-09-19
Documented user registration process

validbeck:

> sc-6057

### Accessing ValidMind
Net-new landing page: [**LIVE PREVIEW**](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-6057/documentation-disclosures-during-user-registration/guide/configuration/accessing-validmind.html)

I also adjusted the guides tile:
Screenshot 2024-09-19 at 11 25 23 AM

### Register with ValidMind
Net-new doc: [**LIVE PREVIEW**](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-6057/documentation-disclosures-during-user-registration/guide/configuration/register-with-validmind.html)

I checked to make sure all the image layouts and such look good on mobile as well!

#### Accept an invite
**This process currently does not allow the user to input profile information or accept disclosures.** See Slack convo here: https://validmind.slack.com/archives/C04FZR7DHMK/p1726767140670419

Made a Story here: https://app.shortcut.com/validmind/story/6614/accept-invitation-to-org-does-not-ask-user-for-profile-t-c-acceptance

#### New `link.register` variable
So we can redirect users to the "Register with ValidMind" page instead of the prod URL: https://github.com/validmind/documentation/blob/29de9ee1fad7301c4925e2c1f71fcb7743e124a3/site/_variables.yml#L38

**I will also need to update the Jupyter Notebooks once the new page is published on `prod`.**


site/guide/configuration/register-with-validmind.qmd
```diff
@@ -0,0 +1,222 @@
...
+::: {.column-margin}
+If you cannot locate this email, ou may need to review your spam or junk mail folders and filters, or request that your IT allow-list our owned domains or retrieve any messages sent by:
+
+a. `validmind.ai`
+b. `validmind.com`
+:::
+
+6. Click **<u>Confirm my account</u>** to gain full access to ValidMind.
+
+    - Return to the main ValidMind dashboard[^16] by clicking on **Back to ValidMind**.
```
nrichers, line 159: Line 159 and 160 feel more like they belong into "What's next" and should not be bullets under 6. Alternatively, they could be regular text that appears after step 6. without any indentation.

nrichers (approved):
This PR provides very thorough coverage — LGTM. :shipit:

For future PRs, I have some guidance from past editors of mine that might be worth discussing further:

- **Make the complex simple**: This topic presents many different options for user registration, along with screenshots and links. From the user's perspective, it might make the registration process feel more involved than it needs to be. If this is a user's first interaction with our content, what could be removed to reduce the feeling of complexity? This isn't a big issue now, but eventually, when we introduce self-service trials, the process will need to feel simple.
  
- **Practice sparse linking**: There are approximately 18 links on this page, some of which are duplicated. That's quite a few. In one of my previous roles, we had a rule that a topic could contain no more than three links, or up to six in exceptional cases. I understand why these links were included — we offer functionality through various providers — but links to tasks performed after registration could be consolidated into the "What's next" section, grouped by role.

- **Add value with annotated screenshots**: Instead of just showing a picture that doesn't explain anything new, add something. For example, in "3. On the page that opens in your web browser...," the screenshot is large, but all the user needs to do is click a button. I would highlight that. You can also reduce the size of these images by using partial screenshots with torn edges, focusing only on the interactive elements.

validbeck:
@nrichers Thank you for taking a look! Made some tweaks to hopefully clean things up a bit:

Screenshot 2024-09-20 at 9 23 38 AM
Screenshot 2024-09-20 at 9 23 46 AM
Screenshot 2024-09-20 at 9 23 53 AM

validbeck:
I think reviewing the screenshots for highlighting and potentially adding a border to ONLY screenshots needs to be a separate Story as it will take some time to test, so I will make one later

PR #428, opened 2024-09-21
Fix staging deploy in Makefile

nrichers:

While fixing the merge conflicts in Beck's PR I noticed that the Makefile had some changes in it from testing that needed to fixed. This PR does just that.


validbeck (approved):
Thank you for catching these!!

PR #423, opened 2024-09-20
Added currency field type to inventory field docs

validbeck:
## Internal Notes for Reviewers

> sc-5944

Revamped "Manage model inventory fields" to accommodate for new currency field types and whenever we get more information about `Calculation` fields.

[**LIVE PREVIEW**](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-5944/documentation-support-for-currency-format/guide/model-inventory/manage-model-inventory-fields.html#inventory-field-types)

PR #420, opened 2024-09-20
Styled screenshots & cleaned up current screenshots

validbeck:
## Internal Notes for Reviewers

> sc-6618

### Style guide: Screenshots

Added a net-new section to reflect our new screenshot styling:

Screenshot 2024-09-20 at 4 08 07 PM

### Cropped & annotated images

- After I applied the `.screenshot` class to applicable images, I cleaned up any that had previous drop-shadowing or "business" around the edges as much as I could
- I also annotated some screenshots that had a specific CTA section

You can see examples on the [**live "Register with ValidMind" preview page.**](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-6618/screenshots-cleanup/guide/configuration/register-with-validmind.html)

### Video formatting

I cleaned these up as well, as I think the rounded corners and drop-shadow are enough to make them stand out and not be too busy:

| Old | New |
|---|---|
| Screenshot 2024-09-20 at 3 16 21 PM | Screenshot 2024-09-20 at 3 37 46 PM |

PR #419, opened 2024-09-20
Preview of Generate with AI docs sections

validbeck:
## Internal Notes for Reviewers

> sc-6619

Mehdi says we are ready to expose the documentation for this feature, **as long as we get approval from @kristof87 for the wording and any other recommendations.** 

### Work with content blocks

This is currently the only page that mentions the feature. 

[**LIVE PREVIEW**](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-6619/discuss-exposing-generate-with-ai-features/guide/model-documentation/work-with-content-blocks.html)

Unhid two sections:

- Stuck on where to start?
- Generate with AI (BETA)

#### Stuck on where to start?

This is in the "Content editing toolbar" overview, links to the Generate with AI (BETA) section below.

Screenshot 2024-09-20 at 11 30 49 AM

#### Generate with AI (BETA) 

This is wrapped under "Add content blocks".

Screenshot 2024-09-20 at 11 30 57 AM

### Other suggestions

Other places I think briefly mentioning the "Generate with AI" feature would make sense:

- Developer Framework > Documenting models > Document models > ["In the ValidMind Platform UI" ` — "For text blocks"](https://docs.validmind.ai/developer/model-documentation/document-models.html)
- FAQ > [Documentation](https://docs.validmind.ai/faq/faq-documentation.html) (@kristof87: Would be nice to get your assist here with writing a question/framing the set up! I can take care of the answer if need be.)

PR #439, opened 2024-09-23
Pushing to `prod` PR#410 onward

validbeck:

- #420
- #423 
- #419


validbeck:
@nrichers No merge conflicts this time but the history between `prod` & `staging` still retains changes from previous to merge: https://github.com/validmind/documentation/compare/prod...staging

Is this working as intended?

nrichers:
> @nrichers No merge conflicts this time but the history between `prod` & `staging` still retains changes from previous to merge: [prod...staging](https://github.com/validmind/documentation/compare/prod...staging)
> 
> Is this working as intended?

How did you merge your previous PR into `prod`? If you squash merged, I think this might be working as designed. The commits from the previous PR would have been squashed down, leaving only a single commit. 

Note that nowhere in our release process do we require you to compare branches to make sure that commits that were merged into main and staging are also present in the PR to merge into prod. I know this sounds counterintuitive, but trying to make comparing git histories part of the release process is not the way to go. 

If you really are curious, I would recommend experimenting — it's good to learn more about how git merges work, but I would create another `prod-test` branch and try out the different options that are available:

image

validbeck:
> If you squash merged, I think this might be working as designed. 

Ah yes, that's what I did. No squash would fix this then? I will try that next time and see what happens. Just want to make sure I understand what's happening under the hood.

nrichers:
> Ah yes, that's what I did. No squash would fix this then? I will try that next time and see what happens. Just want to make sure I understand what's happening under the hood.

"Fix" would imply something is broken — it isn't. You CAN squash merge but it will behave differently than if you created, say, a merge commit. If you squash, the next time you go to merge, GitHub will look at the commit history in `staging` and `prod` and see that there are commit SHAs in `staging` that aren't enumerated in `prod`, so it's going to add them to the list for reference. It has no effect on the changes that get merged in, though, you're still safely merging one branch into another.

Note also that if you change the merge method for one PR in your browser, you might have to remember change it back for the next PR, or you're going to add merge commits to our history in `main`. Again, not something that needs to be fixed, just something to remember. 

For reference, we use `--merge` to create a merge commit [whenever we merge from `main` into `staging](https://github.com/validmind/documentation/blob/main/.github/workflows/merge-main-into-staging.yaml#L44). I did this to preserve the commit history, but if the histories end up becoming an ongoing issue, I might suggest we squash merge into `staging` and sidestep the issue altogether. You can always run a `git diff` on branches instead if you want to know if there are file differences.

PR #443, opened 2024-09-24
Calculation field instructions & example

validbeck:

> sc-5953

### Manage model inventory fields

[**LIVE EXAMPLE**](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-5953/documentation-calculated-fields/guide/model-inventory/manage-model-inventory-fields.html)


validbeck:
@juanmleng Can I get your input on whether or not this is a reasonable example?

Screenshot 2024-09-24 at 11 46 47 AM

If not, can you provide one based on an easy calculation I can use instead?

juanmleng:
The small issue with the current example is that even in the case where both complexity and materiality have large numbers (e.g., low risk), the output could be high risk if complexity is greater than materiality or if both have the same large number (e.g., both low risk, then it would result in high risk).

To make it a bit more realistic, below is a suggestion where materiality and complexity are defined as High Risk or Low Risk, and the risk tier follows a three-level system. 

```
def calculate_risk(params):
    
    # High Risk: If materiality is high risk, return high risk regardless of complexity
    if params.materiality == "High Risk":
        return "High Risk"
    
    # Medium Risk: If materiality is low risk but complexity is high risk, return medium risk
    if params.materiality == "Low Risk" and params.complexity  == "High Risk":
        return "Medium Risk"
    
    # Low Risk: Both materiality and complexity are low risk
    return "Low Risk"
```

validbeck:
@juanmleng Thank you! For some reason our calculated field NEEDS the function name of the formula to be `formula` to be valid, but I have updated the example: 

Screenshot 2024-09-26 at 1 09 51 PM

Screenshot 2024-09-26 at 1 18 23 PM

nrichers (approved):
LGTM 🚀

PR #461, opened 2024-10-02
Nrichers/sc 6760/store python docs as zip file and extract

nrichers:

This PR modifies the `make python-docs` action to replace our folder of HMTL Python docs with a single .zip file that gets unzipped on the fly. This change should stop `git status` from cluttering your view if you run `quarto preview`. 

Easiest way to test, requires that you ran `make get-source` at some point:

- Delete `site/python-docs.zip` and the `site/_site/validmind/` folder.
- Run `quarto render` and check that the `python-docs.zip` file reappears in the site output folder and that the ValidMind Developer Framework reference still works when you open the docs site locally.


validbeck (approved):
Sorry I lost this in my inbox for a while! It looks good, thank you for setting it up!

nrichers:
There was an issue with the paths inside the .zip file which is why the preview didn't work. Fixed in [90a5b0d](https://github.com/validmind/documentation/pull/461/commits/90a5b0d31be5e8c0e457d4f39a511329dfb1cb2e) — now the preview works as expected and the Python docs also work locally.

PR #469, opened 2024-10-09
Docs audit p2 — Get Started

validbeck:

> sc-6525

I cleaned up/clarified instructions under "Get Started", including updating some outdated screenshots: [**LIVE PREVIEW**](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-6525/docs-audit-p2-get-started/get-started/get-started.html)


nrichers (approved):
This is a helpful PR that makes a number of incremental improvements — LGTM! I especially like the next step topic, which is now much more useful for the different roles. :shipit:

Ideas for the future we should chat about:

- It might be helpful to do a fairly ruthless **edit for minimalism** — let's target a 30% reduction in verbiage and see how far we can get. I look at these topics and often think, “this could go, and that, and that…” I'd love for us to add lightness in a future iteration of this content. We could even do a bit of pair programming on one of these topics together to develop a common approach.
- Static screenshots in this section should be replaced with **short, FAQ-style videos**, which we should plan for in a sprint sometime soon. For example, the topic on registering your first model could have a one-minute video at the top that walks through the registration process and then shows where to find the model documentation and validation report.
- We should talk to Andres about retiring Colab, leaving only two QuickStart paths (JupyterHub or cloning the GitHub repo). Colab is chronically out of date, and the process to maintain it is broken. Case in point: at the moment, the version we have in Google Drive for Colab dates from July, but the latest version in GitHub is from September, so they’re a full two months apart.

validbeck:
> It might be helpful to do a fairly ruthless edit for minimalism — let's target a 30% reduction in verbiage and see how far we can get. 

Sounds good, the goal of this 1st round of audits was just accuracy & coherency & adherence to style, not necessarily "full rewrite". Can you make some related Stories encapsulating the objectives you want?

> Static screenshots in this section should be replaced with short, FAQ-style videos, which we should plan for in a sprint sometime soon. 

Same comment as above. :)

> We should talk to Andres about retiring Colab, leaving only two QuickStart paths (JupyterHub or cloning the GitHub repo). 

I can add this to the update the JH instance process, but even that isn't really being upkept. There's also a Story for automating that flow but I don't think it's getting any attention. Do you want to start the convo with Andres as to how we can automate this or remove it?

PR #474, opened 2024-10-10
Applied small suggestions from product offsite

validbeck:

- `min-height` for hero banner so it doesn't get cut off on wider screens
- Removed the extraneous images in the role sections and added some `For` lead-in text


nrichers (approved):
LGTM! 🚀 Thank you for making these changes, looks great in the preview.

PR #486, opened 2024-10-22
Updated tachyons buttons & added video class

validbeck:

> sc-7046

**[LIVE PREVIEW](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-7046/investigate-using-tachyons-css-buttons/index.html)**

- Added a regex search to confirm that URLs starting with `docs.validmind.ai` aren't external links & shouldn't show the icon (like for our developer reference): https://github.com/validmind/documentation/blob/cfa3ba3ea176b91a0c94d2b3415d81b14c45bee1/site/_quarto.yml#L378
- Updated the buttons across the site (even the index page!) to use the markdown links instead of the long terrible HTML links, allowing those links to be rendered by Quarto: https://github.com/validmind/documentation/blob/e25690edf2cc077e5e2864896980b63b7f5481b8/site/styles.css#L383
- Added the same convention to the training & updated buttons in training: https://github.com/validmind/documentation/blob/e25690edf2cc077e5e2864896980b63b7f5481b8/site/training/custom.scss#L16
- Added the rounded corners on screenshots to training as well: https://github.com/validmind/documentation/blob/e25690edf2cc077e5e2864896980b63b7f5481b8/site/training/custom.scss#L9
- Added a video style class & updated any videos to match; added a section to the style guide: https://github.com/validmind/documentation/blob/e25690edf2cc077e5e2864896980b63b7f5481b8/site/styles.css#L453

| Training | Videos |
|---|---|
| Screenshot 2024-10-22 at 4 53 43 PM | Screenshot 2024-10-22 at 5 05 34 PM |


site/index.qmd
```diff
@@ -197,9 +199,9 @@ The **purpose-built platform** for model risk management teams to test, document
 :::{#admin}
 :::
 
-<div style="text-align: center;">
-<a class="btn btn-primary" role="button" style="font-variant: small-caps;padding: 12px;color: var(--bs-white);background: #083E44;border-radius: 4px;margin-top: 4px;font-family: 'Inter';font-style: normal;font-weight: 700;font-size: 20px;line-height: 100%;border: 1px solid #083E44;" href="guide/guides.html">all user guides</a>
-</div>
+<center>
```
nrichers, line 202: Hard "no" on wrapping our new, simpler button in yet more HTML **44 times** in this PR to make things complicated again ... Let's not use HTML when a simpler solution is readily available, please?

In this case, you can add `.tc` to line 189 and `.tl` to line 195 to achieve exactly the same effect with Tachyons. See faccf9191651f60038efec8d63e90523e05a2bb5 for an example on another column to show you what I mean.

nrichers (changes requested):
This PR makes some wonderful changes and I'll approve it as soon as the <center></center> HTML hacks are replaced — we have Tachyons CSS or bog-standard Quarto stuff to let us sort out alignment, no need to add yet more HTML.

The new video class is great, BTW. I have a feeling we will get a lot of mileage out of that as we add more and more videos. And the new rounded buttons and screenshots to also update the training materials are 🔥🔥🔥.

site/releases/2023/2023-sep-27/highlights.qmd
```diff
@@ -306,25 +255,13 @@ We made a number of changes to tests to improve the developer experience:
...
-<form method="get" action="/tests/model_validation/sklearn/ClassifierPerformance.html" target="_blank">
-   <button type="submit" style="color: white; background-color: #de257e; border-radius: 8px; border: none; font-size: 16px; padding: 6.25px 12.5px; margin-left: 16px; margin-bottom: 20px;">ClassifierPerformance
-</button>
-</form>
-
-<form method="get" action="/tests/model_validation/sklearn/MinimumF1Score.html" target="_blank">
-   <button type="submit" style="color: white; background-color: #de257e; border-radius: 8px; border: none; font-size: 16px; padding: 6.25px 12.5px; margin-left: 16px; margin-bottom: 20px;">MinimumF1Score
-</button>
-</form>
+[run_test_suite()](https://docs.validmind.ai/validmind/validmind.html#run_test_suite){.button target="_blank"}
```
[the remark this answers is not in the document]
validbeck: We tried this in the releases cleanup PR, and the relative links didn't render properly, remember?

site/index.qmd
validbeck: For this one this ONLY aligns the buttons — not the top lead-in text or the text for the listings for example. I've already simplified the centering for the tachyons flex-boxes with the `.tc` as much as possible with the exception of when the buttons are part of other text and the other text shouldn't be centered.

validbeck: Ah, I see what you mean here, thank you for the example! I forgot the lead-in text was its own element.

validbeck: All the other instances ARE buttons that are part of a larger body text that isn't their own element but I figured we could wrap them in the tachyons css instead of the HTML.

site/releases/2023/2023-sep-27/highlights.qmd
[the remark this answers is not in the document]
validbeck: I've changed them in the code, but the PR preview links won't work anymore.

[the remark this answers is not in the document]
nrichers: I do remember that discussion but there was an issue with the paths in the python-docs.zip file that I fixed after that (which explained why links would work for you locally but NOT in the PR preview — you happened to have the _source folder locally but the PR previews do not have access to that). I just tested locally and relative links should  work.

site/index.qmd
nrichers: > All the other instances ARE buttons that are part of a larger body text that isn't their own element but I figured we could wrap them in the tachyons css instead of the HTML.

Yes, using Tachyons would work though if we're spending a lot of time centering buttons manually, it might be worth chatting about WHY we're doing this. Authoring should be a simple process that is able to focus on the content and not on individually tweaking the appearance of elements.

site/releases/2023/2023-sep-27/highlights.qmd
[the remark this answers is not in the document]
nrichers: > I've changed them in the code, but the PR preview links won't work anymore.

I won't be able to get to it tonight, but we can do a bit of investigating tomorrow. If necessary, we can download the PR preview via `aws` — stuff that works locally generally should work in the previews or there is a bug somewhere.

For reference, the October 22 release notes used relative links for the Python docs and these work just fine in your PR branch: 

https://github.com/validmind/documentation/blob/beck/sc-7046/investigate-using-tachyons-css-buttons/site/releases/2024-oct-22/release-notes.qmd#L120-L123

In the preview for your PR: https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-7046/investigate-using-tachyons-css-buttons/releases/2024-oct-22/release-notes.html#support-for-bias-and-fairness-tests

(Your links use anchors but that should not matter.)

nrichers:
@validbeck going back to the naming convention for buttons, I started experimenting with course cards for our training pages and this button actually look kind of useful over a light background:

image

I sort of wonder, perhaps we could settle on a more generic naming convention than `{.guides .index}` for the more specialized buttons you created for the home page?

validbeck:
> I sort of wonder, perhaps we could settle on a more generic naming convention than {.guides .index} for the more specialized buttons you created for the home page?

Sure, why don't you play and make the necessary adjustments since you seem to have an idea of what you want? Rather than us just going back and forth! Be the change.

site/releases/2023/2023-sep-27/highlights.qmd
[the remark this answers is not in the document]
validbeck: Looks like it was resolved then, but the other PR had links that worked locally but not in the PR preview.

site/index.qmd
validbeck: These are just one or two per page in the releases only, not in the core guides. Just the exception with all the images and examples being bumped up against each other in the flex-boxes, etc.

nrichers:
> Sure, why don't you play and make the necessary adjustments since you seem to have an idea of what you want? Rather than us just going back and forth! Be the change.

Thanks! I pushed a commit to have a standardized nomenclature for buttons. We now have:

- `.button`
- `.button-pink`
- `.button-white`
- `.button-green`

And then your `.cta` basically applies some special formatting that can be combined with these. The only other thing I did was reshuffle styles.css a bit to group the classes better.

nrichers (approved):
LGTM! This PR has turned out great — I'll pull this into my training updates PR as soon as you merge. :shipit:

PR #501, opened 2024-10-31
Training style overhaul + style/accuracy edit

validbeck:

> sc-7134

[**LIVE PREVIEW**](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-7134/docs-audit-p6-training/training/training.html)

### New training stylesheet

> [site/training/assets/training.css](https://github.com/validmind/documentation/blob/beck/sc-7134/docs-audit-p6-training/site/training/assets/training.css)

Now the training pages have their own adjusted look and feel — the dark green and light teal are the focus instead of the pink: 

| Old | New |
|---|---|
| Screenshot 2024-10-31 at 11 06 00 AM| Screenshot 2024-10-31 at 11 23 44 AM|
| Screenshot 2024-10-31 at 11 06 40 AM | Screenshot 2024-10-31 at 11 23 56 AM|

### New Revealjs slides styling

> [site/training/assets/custom.scss](https://github.com/validmind/documentation/blob/beck/sc-7134/docs-audit-p6-training/site/training/assets/custom.scss)

- Updated too to match the new training styling are some visual adjustments to the slides:

| Old | New |
|---|---|
| Screenshot 2024-10-31 at 11 07 24 AM |Screenshot 2024-10-31 at 11 24 33 AM |
| Screenshot 2024-10-31 at 11 08 15 AM | Screenshot 2024-10-31 at 11 24 43 AM|
| Screenshot 2024-10-31 at 11 26 45 AM | Screenshot 2024-10-31 at 11 26 32 AM|
| Screenshot 2024-10-31 at 11 08 25 AM | Screenshot 2024-10-31 at 11 56 51 AM|
|Screenshot 2024-10-31 at 11 09 13 AM | Screenshot 2024-10-31 at 11 25 10 AM|

- As you can see from the last screenshot, there is now a handy dandy new `.overlay` class with our own custom styling that matches our theme instead of pulling from the Tachyons library (by calling one class rather than multiple additional Tachyons CSS ones!):
https://github.com/validmind/documentation/blob/edbbc2efa5645ad369ed1b7c281a68c7acaf3f29/site/training/assets/custom.scss#L39
- There is also an `.embed` class that takes care of the slide navigation single-source file styling and the callouts within the overlays:
https://github.com/validmind/documentation/blob/edbbc2efa5645ad369ed1b7c281a68c7acaf3f29/site/training/assets/custom.scss#L52

### Reorganized contents of training folder

- It was gettin' lorg. I've moved things into their own folder and adjusted links and references to accommodate the migration:

| Old | New |
|---|---|
| Screenshot 2024-10-31 at 11 12 21 AM| Screenshot 2024-10-31 at 11 20 23 AM |

- There's also a new `fundamentals.qmd` template in our internal templates folder: [internal/templates/fundamentals.qmd](https://github.com/validmind/documentation/blob/beck/sc-7134/docs-audit-p6-training/internal/templates/fundamentals.qmd)

### Slide updates

- There were some broken anchor links out to the old version of the training landing page, which I fixed.
- Made sure the styling/text-distribution for the slides looked a bit more balanced where there were two columns.

#### Administrator fundamentals

I single-sourced the basic workflow setup instructions and inserted them into the `Configure workflows` section, and removed any references to the outdated model documentation workflow that used to exist:

| Old | New |
|---|---|
| Screenshot 2024-10-31 at 11 18 34 AM | Screenshot 2024-10-31 at 11 35 55 AM |

### Role & status styling

- There's a new `.bubble` class to reflect the way roles and statuses are displayed within the UI: 
https://github.com/validmind/documentation/blob/edbbc2efa5645ad369ed1b7c281a68c7acaf3f29/site/styles.css#L470

| Roles | Statuses |
|---|---|
| Screenshot 2024-10-31 at 11 21 54 AM | Screenshot 2024-10-31 at 11 22 01 AM|

- I've updated anywhere that referenced these to use the new style.

| Old | New |
|---|---|
| Screenshot 2024-10-31 at 11 13 06 AM | Screenshot 2024-10-31 at 11 40 23 AM |
| Screenshot 2024-10-31 at 11 16 12 AM | Screenshot 2024-10-31 at 11 40 54 AM |
| Screenshot 2024-10-31 at 11 19 24 AM |Screenshot 2024-10-31 at 11 37 58 AM |


nrichers (approved):
I love this PR. 💖 You made a TON of improvements to our training content that move the needle on the look and feel, introducing that separate-yet-related look we talked about. I also appreciate the many smaller changes you made, like the use of `{.bubble}`. 

My only suggestion would be around the sidebar on the main training page which we talked about a bit already. I agree with your assessment about the look and feel being off if you just use the regular background but my question is: do we NEED a sidebar on that page at all? 

I pushed d19bd85b33584cad983deb041b2aa6ad614b46e9 to show what a sidebar-less training landing page might look like. The only useful thing you lose is the search box but with the reimplementation of search on the front page, we could move search into the navbar.

validbeck:
It still didn't look quite right without the sidebar (so much empty space, and using the page fill options made it worse...), so what I did instead was get rid of the navbar to reduce visual clutter: 

Screenshot 2024-10-31 at 5 30 00 PM

nrichers:
> It still didn't look quite right without the sidebar (so much empty space, and using the page fill options made it worse...), so what I did instead was get rid of the navbar to reduce visual clutter

Can we wait with merging this PR until we've had a chance to make some additional changes, please? I think I had revelation while thinking about this and looking at some other training portals for inspiration.

Eventually, that training landing page will morph into something that stands alone as the home page on a separate training portal. That new training portal will feel more like the current home page of our docs site, with: 

- Its own hero banner
- Sections with course tiles
- Highlights or video links
- And much more!
 
In preparation for that future, my preference would be to get rid of the sidebar long before getting rid of the navbar. I'd like there to be some natural progression as we iterate over the training content. 

So here's the revelation, though you kinda have to squint and imagine the future ... For now, maybe the answer lies in introducing another item of visual interest, like a hero banner that stretches across to solve the issue you are concerned about. That would solve the empty space problem as the banner can stretch across, and then we list the current content as is. 

Take a look at [Airtable's academy](https://academy.airtable.com/) for an extremely simple example. And I'll even let you make our hero banner somewhat pink — if the docs home site uses pink as the main accent colour with a dark green hero banner, then the training site could invert this colour scheme.

nrichers:
Related to my previous comment, I'd like to discuss the new training content layout with Mehdi this coming week. It would be helpful to have that precursor to a new training portal home page in place, even if it requires some squinting and imagination for now.

validbeck:
> Can we wait with merging this PR until we've had a chance to make some additional changes, please? I think I had revelation while thinking about this and looking at some other training portals for inspiration.

It's why I left this as is without merging, yes.

I personally really like the sidebar as opposed to the navbar (and I actually based the entire redesign around it, so to me it doesn't feel cohesive without it), but we'll just have to agree to disagree there and you do happen to have an override if you're really adamant about it. :) 

> For now, maybe the answer lies in introducing another item of visual interest, like a hero banner that stretches across to solve the issue you are concerned about. 

It's not really about "stretching across" for me (as I played around with scretching the page content before I readded the sidebar)... it's just that I think the swap between the nav & sidebar really visually distinct and I like that much more.

nrichers:
> I personally really like the sidebar as opposed to the navbar (and I actually based the entire redesign around it, so to me it doesn't feel cohesive without it), but we'll just have to agree to disagree there and you do happen to have an override if you're really adamant about it. :)

I don't think it works for a training home page and there's a reason we don't use this approach on our home page. But let's let this sit for a few hours and discuss later. I need to focus on some product footage for Kevin for now and won't have time to tinker with alternatives until the afternoon most likely.

nrichers:
> ### Reorganized contents of training folder

Another item for us to discuss: I like the idea of breaking things down into different paths, but the folder path should reflect the course name. Soon enough, we will have multiple courses for developers, for example, so rather than having to rename existing courses, it would be good to agree on a convention now that does not require future file reshuffling. This will become especially important as we will likely have quite a few links from our CSM org and other places to training content.

PR #509, opened 2024-11-05
Training course templates & organizational refinement

validbeck:

> sc-7323

## Course folder renames

| Old | New |
|---|---|
| administrator | administrator-fundamentals |
| developer | developer-fundamentals |
| validator | validator-fundamentals |
| wip | risk-governance-fundamentals |

## Training templates

There's now a new `training-templates` folder that lives within the `/training` directory. **These pages are set to `search: false`.**

- `course-registration`serves as a basic template for registration/course outline pages: [**LIVE PREVIEW**](url)
- `course-slides.qmd` captures our commonly used slide styles in a live demo: [**LIVE PREVIEW**](url)

## Style guide rework

There was a random preview section that wasn't nested under the training materials, so I've just stolen some of the content from that and reworked the "Training" section:

 [**LIVE PREVIEW**](url)

## Other edits

- Trying out "ValidMind Academy" for the training section, since it's its own neat semi-separate portal now to align with how we talk about it in the rest of the site.
- `custom.scss` > `slides.scss` to match our style guide rule of filenames should reflect what they're for. 
- There were some html centers hanging around in the training slides and some awkward lead-in text; I've fixed those. 

### Video templates?

These are still using the old slide sliding/site theme. I left them untouched as I don't know if we wanted to update these, given that videos are being filmed as we speak...


site/_quarto.yml
```diff
@@ -322,11 +322,11 @@ website:
     #     - text: "---"
```
nrichers, line 322: I think we can delete these commented-out lines? We don't need them anymore.

nrichers (approved):
LGTM! (Looks GREAT to me). 🚀 You continue to make all of these incremental improvements to our site that add up to a far nicer look & feel. The new template for training content is a nice touch and will come in very useful — I might even co-opt it for part of our tech demo.

validbeck:
@nrichers 

I simplified the styles for the default listings so that they display like this everywhere but training (the training will universally have the green styled table as long as the page is calling from `training.css`):

Screenshot 2024-11-05 at 4 56 52 PM

nrichers:
> I simplified the styles for the default listings so that they display like this everywhere but training (the training will universally have the green styled table as long as the page is calling from `training.css`):

Very nice! This simplified style seems to work well. 

For the external URLs you use in your example, I currently use the .preview extension but we can perhaps chat tomorrow about what makes the most sense.

PR #273, opened 2024-08-14
Updated permissions in docs

validbeck:
## Internal Notes for Reviewers

> For sc-4995, I:
> - Established a new `prereq` convention & [CSS class](https://github.com/validmind/documentation/blob/d2c4af5691ab5cd21281dc315b2fe3508bbcf736/site/styles.css#L267)
> - [Added a variable back in for our login information](https://github.com/validmind/documentation/blob/d2c4af5691ab5cd21281dc315b2fe3508bbcf736/site/_variables.yml#L37)
> - Added a framework in `configuration/manage-permissions.qmd` for permissions tables (when the correct permissions are released)
> - Adjusted any prereq blurbs to be more inclusive of permissions/roles, linking out to context when relevant
> - Did a preliminary review of modified instructions for accuracy (removed repetitive "1. Log in to ValidMind," cleaned up inaccurate references, etc.). **[See comment on audit Story](https://app.shortcut.com/validmind/story/4902/docs-site-audit-accuracy-tone-review) for remaining articles to review.**

Whenever permissions change or are added, only one page has to be updated (more or less). 

permissions-table

### Managing users

The `Supported Roles` section is now `Default roles`, to accommodate for the ability to add custom roles and toggle granular permissions:

| Old | New |
|---|---|
| Screenshot 2024-08-14 at 9 35 09 AM | default-roles |

- While it is called the `prereq` class, it's just some styling — I've also used it in a few other places to draw attention to the central information on that page.
- It's just an FA icon in front of the header text, so it can be changed depending on the section.

| Old | New |
|---|---|
| Screenshot 2024-08-14 at 9 46 08 AM | Screenshot 2024-08-12 at 3 01 47 PM |
| Screenshot 2024-08-14 at 9 47 14 AM | Screenshot 2024-08-12 at 3 02 37 PM |

#### Margin styling

Depending on whether or not there are margin footnotes, the section will span the entire body + margin or just the body:

- I personally rather like this effect, since if there are margin footnotes the continuation of the section draws the eye to the supplementary information in a subtle way.
- If there aren't any footnotes, it keeps the line of sight contained to the body only.

| Margin footnotes | No margin footnotes |
|---|---|
| Screenshot 2024-08-12 at 3 03 36 PM | Screenshot 2024-08-12 at 3 05 43 PM|

#### Prerequisites
These are now a task list:

- This solves the problem of the login-link variable adding extraneous spacing before and after the variable line when in a bulleted list, causing the other items needing to be spaced.
[...]

PR #510, opened 2024-11-06
Document LLM features

nrichers:

This PR includes information about:

- Our philosophy on using LLMs
- A short explanation of the four features
- Information about accessing our policies
- Requesting a demo
- Related screenshots

I'd be interested to hear if we should change the title to something that has "GenAI" or "AI" in it. 

 [**LIVE PREVIEW** 🧶](https://docs-demo.vm.validmind.ai/pr_previews/nrichers/sc-6909/document-model-documentation-for-validmind/about/overview-llm-features.html)

## External Release Notes

{{< var vm.product >}} offers several specialized features that use large language models (LLMs) to streamline model risk management and ensure regulatory compliance. Here's how we approach these features and what you need to know: [Large language model features](/about/overview-llm-features.qmd)


validbeck:
@nrichers Played around with the styling some.

- New `.feature` class for a different type of callout.
- I still think the grid cards work better for the `What's next` section as opposed to the table. We use the card format as a CTA across the site, so I personally thinks it just looks a bit more uniform. The other option would be to do some columns with some CTA `.button`s like we do in our releases...

validbeck (approved):
Content looks good to me! 🎉

site/about/overview-llm-features.qmd
```diff
@@ -0,0 +1,113 @@
...
+    image-align: left
+    contents:
+    - path: https://validmind.com/about/legal
+      title: Read our legal ease
+      description: We believe in transparency across all our practices — not just because it’s essential to earn your trust, but because it’s the right thing to do.<br><br>Understanding our policies shouldn’t feel like deciphering code, so we’ve made our legal texts as clear and accessible as possible. These documents cover the details of our practices, including data handling, privacy, responsible AI use, and regulatory compliance.
+      # author: validmind.com
+    - path: https://validmind.com/contact/
+      title: Request a demo 
+      image: /assets/img/hero-platform.png
+      description: Discover how ValidMind’s LLM-powered platform, purpose-built for model risk management teams, enables fast, confident testing, documentation, validation, and governance of Generative AI, AI, and statistical models.
```
validbeck, line 23:
```suggestion
      description: Discover how {{< var vm.product >}}’s LLM-powered platform, purpose-built for model risk management teams, enables fast, confident testing, documentation, validation, and governance of Generative AI, AI, and statistical models.
```

site/about/overview-llm-features.qmd
```diff
@@ -0,0 +1,113 @@
...
+
+::: {.attn}
+## {{< fa list-check >}} Our philosophy
+
+<!-- Don't add a caption, it will push the footnote outside the section -->
+::: {.column-margin}
+![](overvierw-llm-features-check-document.png){fig-alt="A screenshot showing the LLM feature for checking documents that can be accessed with the Check Document button" class="screenshot"}
+:::
+
+At ValidMind, our approach to AI features, particularly in testing and risk management, reflects a philosophy of _drinking our own champagne_ or using our own products. This means that, while we create risk management tools for our users, we also rigorously apply these tools and practices within our own product development.
```
validbeck, line 38:
```suggestion
At {{< var vm.product >}}, our approach to AI features, particularly in testing and risk management, reflects a philosophy of _drinking our own champagne_ or using our own products. This means that, while we create risk management tools for our users, we also rigorously apply these tools and practices within our own product development.
```

validbeck (commented):
Missed a few product variables. Though I'm torn on whether or not we want to use this one, as I go and replace them all within our product renaming... hm.

nrichers:
> * New `.feature` class for a different type of callout.

I quite like the idea of behind this change but are you able to move the checkmark graphics to the left of the text? I find the overlap of text and graphics to be much harder to read than the original. 

image

> I still think the grid cards work better for the What's next section 

Ha. After our convo yesterday and after looking at your other PR example, I decided to ditch the URL preview I had in the early versions of this PR. I still maintain that external links could just use the `.preview` option rather than require manually adding some images for listings, especially as those external URLs have good tag lines on them that show up. 

> Missed a few product variables. Though I'm torn on whether or not we want to use this one, as I go and replace them all within our product renaming... hm.

How about this? Whichever PR gets merged first adopts the content/convention from the other? E.g. if you merge your PR first, I will update mine; if I merge first, we can pull this content into yours. Happy to help.

nrichers:
@validbeck one other question for you: `.attn` in the training CSS seems to be main column width but here it appears to be page width which can look a bit funny in higher resolutions. I can reverse engineer this CSS but you probably remember where this is defined much better. Would you able to update `.attn` to match how the training CSS behaves?

image

nrichers:
> Like so?

Yes, that looks much better, thank you! Having the checkmarks appear on the same level as the text would be a further improvement but this already looks much cleaner. We don't really have a visual standard for referring to features so unless @kristof87 has a preference here, we could just leave this as-is. 

> We talked about this quirk here (and had several conversations about how it's difficult to get a workaround via CSS, since Quarto considers the margin/footnotes to be part of the div, and very rarely do people maximize to what you do!): https://github.com/validmind/documentation/pull/273 (CTRL + F for "Margin styling".)

Yes. I do remember this conversation but perhaps I just don't understand the CSS limitations here — this _should_ be solvable and the current behavior feels like a bug. 

Based on what I see in https://quarto.org/docs/authoring/article-layout.html#overflowing-content, it looks like we're applying something like `.column-screen-right` rather than `.column-page-right`, but in CSS. 

Further, the behavior described in the PR you refer to under "Margin styling" is NOT actually how it behaves in real life ("the section will span the entire body + margin or just the body"). 

You don't even have to go to ridiculously high resolutions to encounter this issue. 4K screens are very common and the page just looks off at that resolution. The same is true for 1440p, although not as severely.

site/about/overview-llm-features.qmd
nrichers: Couldn't commit this directly for some reason but fixed in the source.

validbeck:
> Yes. I do remember this conversation but perhaps I just don't understand the CSS limitations here — this should be solvable and the current behavior feels like a bug.

To me it's not a bug, like I mentioned in our original conversations. To me it says "these footnotes are part of this div"; it draws your eyes to the fact that they are connected.

For example, if you look at it in mobile view, they are wrapped into the section:

Screenshot 2024-11-06 at 2 13 04 PM

> EDIT: It looks like we're applying a couple of things here that are potentially problematic and most likely it's page-full. The thing is, I don't know where they are coming from. 

This isn't what's causing it. The training pages, like you mention, have the `page-full` explicitly because I wanted to expand the section to cover up the white space left by the hidden sidenav. But they don't do the footnotes thing because the `.attn` section doesn't have footnotes. Again, this is a feature and not a bug.

It IS something Quarto is applying behind the scenes, PRECISELY because there are footnotes that are nested in that `.attn` div. 

> Further, the behavior described in the PR you refer to under "Margin styling" is NOT actually how it behaves in real life ("the section will span the entire body + margin or just the body").

I feel like this might just be a pedantic nitpick at this point? 😅 I get what you're saying, in that what Quarto actually does is force the content to span the entire body past the margin, but what it naturally considers the "body + margin" is the entire right half of the page in that situation. When there is content in the margin, it auto-applies some styling (squishes the content to be narrower, for one, then wraps everything into whatever div is active lengthwise). 

**If it is a "bug", it's a Quarto level bug, not something we introduced with the styling. We'd have to hack it at the Quarto rendering level, not the CSS level.** Feel free to prove me wrong, but I spent a few hours fiddling with it and the version we have is the best we got so far. For example...

```
.attn.page-columns.page-full {
  width: 1500px;
}
```

Jankifies it, even though it limits the `.attn` section to the width specified: (also looks awful on mobile, no longer responsive)

Screenshot 2024-11-06 at 2 37 17 PM

Forcing a page-layout of "article" also doesn't work — we'd have to hack together an even more annoying `custom` layout (if that would even fix the issue) for each page you want to go down that route... Adjusting the `margin-width` only affects the margin contents, etc...

> I also add the option to offset URL previews, so something like this is possible which looks very clean:

Oh, that does look OK. I am all right with this version, it was just the previous version with the navbar that really looked disjointed.

site/styles.css
```diff
@@ -507,8 +507,8 @@ figcaption {
   text-shadow: 2px 0 #fff, -2px 0 #fff, 0 2px #fff, 0 -2px #fff, 1px 1px #fff, -1px -1px #fff, 1px -1px #fff, -1px 1px #fff;
   background-color: white;
   background-image: url('/assets/img/ecosystem-checks.svg');
-  background-size: 40%;
-  background-position: right 0px bottom 0px; 
+  background-size: 30%;
```
validbeck, line 510: I don't like it at this size — I tweaked it pretty much all the way from 10% to 100% and the 40% was the sweet spot where the faint green checkmark was still visible.

validbeck:
Why is there weird whitespace on the right in the previews? 🤔 

Screenshot 2024-11-06 at 2 51 26 PM

site/styles.css
validbeck: Also, now you've moved it around so much there is too much padding at the bottom of the `.feature`.

nrichers: I reverted the size to 40% but had to increase the offset a bit more to avoid overlapping with the text. This works for me if you're OK with it — the original asymmetrical or diagonal placement of the checkmarks looked off to me, hence the change. I actually quite like how things look now, not too subtle, not too distracting. 

image

nrichers:
> Why is there weird whitespace on the right in the previews? 🤔

Interesting, I do not see that in my local preview. It's possible the horizontal Tachyons CSS directive for padding is affecting this, will look into it. Here's what I see:

image

site/styles.css
validbeck: It kind of just looks like you've cut the light-green checkmark off now? 😅  (The original version was WAY flashier than any of commits I send to the cloud, so just know I toned it down for you)

site/about/overview-llm-features.qmd
```diff
@@ -138,4 +138,6 @@ Discover how {{< var vm.product >}}’s LLM-powered platform, purpose-built for
 
 <!-- FOOTNOTES -->
 
-[^1]: [Run tests and test suites](//developer/model-testing/testing-overview.qmd)
\ No newline at end of file
+[^1]: [Run tests and test suites](//developer/model-testing/testing-overview.qmd)
+
+[^2]: [AI Usage Policy](https://validmind.com/about/legal/ai-usage-policy/)
```
validbeck, line 143: If you have the direct link then this footnote isn't needed — can we pick one or the other?

```
[WARNING] Note with key '2' defined at line 122 column 1 but not used.
```

nrichers:
> **If it is a "bug", it's a Quarto level bug, not something we introduced with the styling. We'd have to hack it at the Quarto rendering level, not the CSS level.** Feel free to prove me wrong, but I spent a few hours fiddling with it and the version we have is the best we got so far. For example...

Actually, I think you put your finger on it ... Either an image in the margins OR a footnote will trigger this behavior. I noticed some janky behaviour with footnotes in responsive columns as well where they will mess up (very technical, I know) the layout. I had to use an inline link to work around this issue. 

My apologies to you Beck. Your changes that introduced `.attn` are not the issue. I appreciate the time you took to explain this further. 

For now, I've worked around this issue by moving the footnote out of the "Philosophy" section and moving the image into a column inside of `.attn`. The end result actually looks decent:

image

BTW, I don't know if you noticed this, but `lightbox: auto` enables you to have popups for images that let you see the full size. This is the first time we've used it as far as I can tell, but it's helpful with images that have a lot of detail, like the ValidChecker.

site/styles.css
nrichers: Let's chat when we meet in 7 mins ... 😁 I'm sure we can figure it out. The light green checkmark being cut off is OK-ish, I think. Would be solved by sizing to 30%! 😛

site/about/overview-llm-features.qmd
nrichers: I already removed this in a commit not too long ago.

site/styles.css
nrichers: Reverted to 40%.

validbeck:
> My apologies to you Beck. Your changes that introduced .attn are not the issue. I appreciate the time you took to explain this further.

I do agree that *ideally* Quarto would not apply the dang styling to the full right half whenever there is margin content, and instead just what they call the body + margins, but for some reason anytime there is margin content it decides to go wild. It doesn't bother me too much (because my screen is not large enough at maximize for me to care), but it might be worth it to log a ticket with the Quarto peeps? 

> BTW, I don't know if you noticed this, but lightbox: auto enables you to have popups for images that let you see the full size. This is the first time we've used it as far as I can tell, but it's helpful with images that have a lot of detail, like the ValidChecker.

I did note that, and I don't think it needs to be on for both images though I don't know why GitHub put my comment for that somewhere else instead of as a review comment. 💀

site/about/overview-llm-features.qmd
nrichers: Yup, removed. The remaining footnote is now commented out but might remove it as well.

nrichers:
> I did note that, and I don't think it needs to be on for both images though I don't know why GitHub put my comment for that somewhere else instead of as a review comment. 💀

It's enabled at the page level, so it's either all or nothing. The animated GIF is kinda hard to read without the lightbox treatment and so ...

validbeck:
> It's enabled at the page level, so it's either all or nothing. The animated GIF is kinda hard to read without the lightbox treatment and so ...

You don't need to enable it at the page level. For example:

```
![Document checker in progress](overview-llm-features-check-document-in-progress.gif){fig-alt="A screenshot showing the LLM feature for checking documents that can be accessed with the Check Document button" class="screenshot" .lightbox}
```

kristof87 (approved):
This looks great. I really like how you are presenting this content.

PR #521, opened 2024-11-12
Updated product terminology — `documentation` repo side

validbeck:
## Internal Notes for Reviewers

> sc-6647

[**LIVE PREVIEW**](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-6647/update-product-terminology/index.html)

### Major changes
Screenshot 2024-11-12 at 3 48 34 PM

- Updated the variable file to reflect new product names & updated the style guide to reflect some additions: 
https://github.com/validmind/documentation/blob/b1e20d51703b0bb3485e9332f2f357926ba3c61e/site/_variables.yml#L16
- Updated any product images that had old terminology:
     - [site/about/fine-print/overview-api-integration.jpg](https://github.com/validmind/documentation/blob/b1e20d51703b0bb3485e9332f2f357926ba3c61e/site/about/fine-print/overview-api-integration.jpg)
     - [site/get-started/validmind-lifecycle.jpg](https://github.com/validmind/documentation/blob/b1e20d51703b0bb3485e9332f2f357926ba3c61e/site/get-started/validmind-lifecycle.jpg)

- Added an auto-uppercasing function in our css so that `_quarto.yml` sidenav headers are able to make use of our product variables & still retain the styling:
https://github.com/validmind/documentation/blob/b1e20d51703b0bb3485e9332f2f357926ba3c61e/site/styles.css#L529
- Developer Framework dropdown > `Developer`:
https://github.com/validmind/documentation/blob/b1e20d51703b0bb3485e9332f2f357926ba3c61e/site/_quarto.yml#L39
- New [`internal/templates/jupyter-notebooks` directory with `.ipynb` files](https://github.com/validmind/documentation/tree/b1e20d51703b0bb3485e9332f2f357926ba3c61e/internal/templates/jupyter-notebooks)
- Proofreading of site for awkward phrasing

#### File renames

Aliases have been set up for the old URLs:

| Old | New |
|---|---|
| `site/developer/get-started-developer-framework.qmd` | `site/developer/get-started-validmind-library.qmd` |
| `site/developer/model-documentation/install-and-initialize-client-library.qmd` | `site/developer/model-documentation/install-and-initialize-validmind-library.qmd` |

### To do

- [ ] Instances of the `developer-framework` GitHub repo URL: **Waiting on the repo rename**
- [ ] Instances of references to the `developer-framework` repo by name: **See previous item**
[...]

PR #524, opened 2024-11-13
New sidebar & ToC theme

validbeck:

From https://github.com/validmind/documentation/pull/521

> Looking at preview, I think we'd need to apply this to both the sidebar navigation and right-hand ToC (or neither) for a consistent feel.

[**LIVE PREVIEW**](https://docs-demo.vm.validmind.ai/pr_previews/beck/sidebar-fiddling/index.html)

| Old | New |
|---|---|
| Screenshot 2024-11-13 at 10 59 03 AM| Screenshot 2024-11-13 at 10 58 45 AM|


nrichers:
If you do want to do a solid background for the sidebar, I'd recommend something subtler that is closer to what the product does: 

image

It's quite subtle. With a solid background, maybe NOT doing the same on the left and the right would be preferable, as the preview feels a bit busy to me. (With JUST a text colour change, I'd try the same on both sides.)

nrichers:
BTW, this is an example of a simple but interesting box treatment: https://graphacademy.neo4j.com/courses/gds-product-introduction/

All the same, I'd use a box or solid background on only one side.

validbeck:
Why do you hate colour so much, Nik?? 😭 😭 😭 

I stole some other greens and some hover states from the platform, and I'm willing to live with this if you are:

Screenshot 2024-11-13 at 3 41 06 PM

Adjusted the drop-down as well since the hover state for the sidebar doesn't play nicely with BG shifts:

Screenshot 2024-11-13 at 3 44 00 PM

Simplified training to one `.css` sheet (and updated the conventions to match):

| Landing | Registration Pages |
|---|---|
| Screenshot 2024-11-13 at 3 44 32 PM | Screenshot 2024-11-13 at 3 41 37 PM |

nrichers (approved):
Nik apologizes to Beck for annoying them and expresses his gratitude and thanks for this PR. :shipit:

PR #528, opened 2024-11-14
Live execution of `intro_for_model_developers_EXECUTED.ipynb`

validbeck:

> sc-7406

### ValidMind Library updates

This PR pulls in the latest changes from updated [`validmind-library`](https://github.com/validmind/validmind-library) via make get-source. I can confirm that the edits show up in the rendered files. 

For example: **[LIVE API REFERENCE PREVIEW](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-7406/replace-executed-intro-for-model-developers/validmind/validmind.html)**

### Live execution of intro_for_model_developers_EXECUTED.ipynb

I've used [Quarto's native JupyterLab](https://quarto.org/docs/tools/jupyter-lab.html) functionality to render this notebook on the fly, and it only cost me some of my sanity — what a bargin.

- Now, the executed notebook RENDERS THE MISSING WIDGET TEMPLATES which are now fully interactive!!!
- It is also now set to [`page-layout: full`](https://quarto.org/docs/output-formats/page-layout.html#page-layout) to cover more of the negative-space left by the missing sidenav. 

| Old | New |
|---|---|
| Screenshot 2024-11-15 at 12 58 13 PM| Screenshot 2024-11-15 at 12 57 40 PM|
| [OLD VERSION](https://docs.validmind.ai/notebooks/tutorials/intro_for_model_developers_EXECUTED.html) | [**NEW LIVE PREVIEW**](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-7406/replace-executed-intro-for-model-developers/notebooks/tutorials/intro_for_model_developers_EXECUTED.html) |

_How does this work?_ Well: 

#### Global Jupyter Notebook execution settings

In `_quarto.yml`, there are now some global `execute` settings determining behaviour for what Quarto should do when it renders `.ipynb` files:

https://github.com/validmind/documentation/blob/54e91758065f3e1c24ef2e9fa6ee27a457627ed2/site/_quarto.yml#L5

- Execution for notebooks is disabled by default
- Notebooks shouldn't be re-executed during re-renders 

#### intro_for_model_developers_EXECUTED.ipynb [front matter](https://quarto.org/docs/tools/jupyter-lab.html#yaml-front-matter)

There is now a `raw` cell at the top of this version of the notebook that determines the granular execution behaviour for rendering this file:

Screenshot 2024-11-15 at 2 00 01 PM

- This front matter is hidden on the rendered version
- Execution for notebooks is enabled
- [Notebook should be re-executed on re-renders](https://quarto.org/docs/projects/code-execution.html#freeze) 

#### Local output for executed notebook added to `.gitignore`

If you preview or render the files locally, this notebook will execute and we don't want the output cruft (since the workflows take care of the execution for our hosted sites):

https://github.com/validmind/documentation/blob/54e91758065f3e1c24ef2e9fa6ee27a457627ed2/.gitignore#L15

- Quarto rendered version of the execution ignored: [`site/_freeze/`](https://quarto.org/docs/projects/code-execution.html#using-freeze)
- Actual output of that particular notebook ignored: `site/notebooks/tutorials/my_tests/`

#### Secure `api_key` & `api_secret`

Screenshot 2024-11-15 at 2 23 06 PM

##### New "Repository secrets"

- For use in the updated GitHub workflows, [two (2) new secrets have been added](https://github.com/validmind/documentation/settings/secrets/actions):
 1. `PLATFORM_API_KEY`
 2. `PLATFORM_API_SECRET`
 - These secrets are linked to this ValidMind Academy model: `cm3j9lu5208mf21ifij2tp6bk`
Screenshot 2024-11-15 at 2 01 21 PM
 - Obviously, the code in the notebook references both the GH workflow secrets and your local `.env` secrets so that it works in both environments — see below.

##### `.env.example`

- Locally, I also have the secrets added to my `.env` file. **WHEN YOU RUN THE NEXT PREVIEW OR RENDER LOCALLY, you'll need to replace this with the Academy org's model's key and secret so that the notebook executes properly, as well as add yourself as a developer to the model, etc.**
- I've updated the `.env.example` as well:

https://github.com/validmind/documentation/blob/54e91758065f3e1c24ef2e9fa6ee27a457627ed2/.env.example#L3

#### Updated GitHub workflows

Screenshot 2024-11-15 at 1 30 45 PM

Some new jobs have been added to all three of our "render docs" workflows:

- `Install python3 for Jupyter Notebooks`: This installs the version of `python3` required for ValidMind & upgrades `pip` for use in `Check dependencies`.
- `Install validmind for notebook execution`: This installs all the `validmind` packages and dependencies required to run the "Intro to Model Developers" notebook with minimal warnings and errors. (**There are a few weird ones,** I went on quite a journey.)
- `Check dependencies`: This installs `pycairo` required for `check` and then runs the dependency check.  

The following job has been updated to include the repository `env` secrets:

- `Render docs site`:

https://github.com/validmind/documentation/blob/cb61d671da6a9bfc40f87940d537befae9cc1a97/.github/workflows/validate-docs-site.yaml#L50

### n.b. 

This PR won't be merged until https://github.com/validmind/documentation/pull/529 is approved and merged as they are intertwined.

#### Local execution of `intro_for_model_developers_EXECUTED.ipynb` 

**EDIT:** We discussed this and we will add this officially to the `.gitignore` in a follow-up PR so we can just leave this file untouched and not override it accidentally with local changes:

Screenshot 2024-11-15 at 4 28 35 PM

- You'll either have to add the Academy model's valid key and secrets as the model's developer so that the notebook renders/previews fine locally (as long as you have most of the packages/dependencies sorted..... [which they might not be](https://www.notion.so/validmind/Documented-Issues-Troubleshooting-bc9a38fc9ab34afd88251a92265c5a6f?pvs=4) (WARNING: 🐰 🕳️ ))
- OR, comment out the following in the front matter and temporarily add this file to your `.gitignore` so that the changes don't get sent up to remote:

```yaml
    code-tools: true
    include-after-body: 
      - https://unpkg.com/@jupyter-widgets/html-manager@*/dist/embed-amd.js
execute: 
  enabled: true
  freeze: auto
```

#### `SHAPGlobalImportance.py` error

- This test is returning some errors but the rest of the tests are still being logged to the platform. 
- Same error under two sections:
1. "Run the model evaluation tests"
2. "Viewing and updating the configuration for the entire model documentation template > Update the config" 
```bash
ERROR(validmind.vm_models.test_suite.test): Failed to run test 'validmind.model_validation.sklearn.SHAPGlobalImportance': (TypeError) loop of ufunc does not support argument 0 of type float which has no callable rint method
```
- This didn't happen the last time I executed the notebook and I spent about half a day fiddling with dependencies and script edits to no avail, so I've just logged a Shortcut Story on advice of Andres: [**sc-7456**](https://app.shortcut.com/validmind/story/7456/intro-for-model-developers-ipynb-throws-shapglobalimportance-py-error)
- The good thing is, since the notebook is being executed on the fly, in *theory* if it is an issue with the Library or the test, once those updates are pulled in the errors in the rendered notebook should resolve (I *think*). 

#### Developer Fundamentals iframe

- This is still linking to the TOP-LEVEL docs-demo rendered version, because for some reason the rendered relative links in the iframe are finicky. This feels similar to the issue with the original relative link buttons in the release previews for the python docs. 
- BUT, since the executed/rendered version is fine outside of that relative link in the preview, it should likely be fine once we send the new executed notebook up to `staging` and `prod`.
- You can inspect this yourself by going into the training, and then on the first slide that has the relative executed notebook as the background, click `View Frame Source`:
Screenshot 2024-11-15 at 1 53 37 PM

#### Added render run-time

Yeah, I know. I'm sorry. For local versions, you can use the `.gitignore` hack for the specific notebook to get around it. I think this is pretty cool functional trade-off though. 

#### Edit: Variables are messed up on mobile view & breadcrumbs

This applies to titles and sidebar links in `_quarto.yml`:

Screenshot 2024-11-19 at 10 40 55 AM

Applied a workaround for now (just using the plain-text term).


nrichers (commented):
This PR is both a great leap forward and problematic in its current form. I'll explain, and then we can figure out how to best incorporate your amazing work here, @validbeck. 🥳 

**TL;DR:**  
My requested changes are:  
- Split the execution into its own workflow that triggers only when the notebook changes
- Remove the `_EXECUTED` notebook altogether by switching execution to a project profile
- Keep the API key and secret where you've added them for future reuse in other projects

Now for the details ...

## A great leap forward

The ability to access a code snippet via a couple of keys enables several interesting possibilities. For example, we could run a code cell in a Jupyter notebook that connects to ValidMind, then surface that code cell in the documentation.

This feature could be used on the front page of the docs to render a Pearson correlation matrix or other eye-catching graphs, instead of relying on static images from our main site, for example.  

Effectively, this PR also enables embedding something like the tech demo into the docs site, which could radically improve the QuickStart or onboarding journey. It could surface parts of product functionality to guide users more effectively.

In short, **this is way cool!**

## Kind of problematic

While this PR accomplishes a lot, it introduces some issues:  
- It solves only some of the problems with the notebook  
- The infrastructure cost is too high  
- There’s unnecessary complexity in the workflow

Let’s take these in turn—and please let me know if I’ve misunderstood anything.

### Solves only some issues with the notebook

While execution may fix JavaScript widgets, several important cells still don’t run in your PR preview, such as:

- `vm.preview_template()` — important in the model development process and repeatedly referenced
- `vm.tests.run_test()` — important to show what ValidMind can do

Additionally, some cells can *never* run in this notebook as-is, because they depend on the user performing actions on the ValidMind platform first. For example:

```
2024-06-12 16:28:18,447 - WARNING(validmind.vm_models.test_suite.runner): Config key 'my_test_provider.ConfusionMatrix' does not match a test_id in the template.
    Ensure you registered a content block with the correct content_id in the template
    The configuration for this test will be ignored.
```

Here, the notebook expects a test-driven block for a confusion matrix to be inserted into the model documentation on the ValidMind Platform. This dependency highlights a limitation with executing a notebook where core functionality relies on external actions outside of the notebook.  

To bypass this, we’d have to manually insert the content block into the model documentation and ensure it remains untouched — adding another dependency we need to track. 

### The infrastructure cost is too high

Rendering one notebook imperfectly now increases the CI check runtime from ~5 minutes to 13 minutes — 3 minutes for environment setup and 9 minutes for rendering. This slowdown impacts every commit as far as I can tell, which is not sustainable.

This additional overhead is unnecessary. Instead of rendering the notebook in the regular workflow, it could be executed only when the file itself changes. A separate workflow could handle this, triggered by changes to the notebook file:

```yaml
on:
  push:
    paths:
      - path/to/notebook.ipynb
```

Since the notebook doesn’t need to appear in the sidebar, we can execute it after the docs site render, copy the output to the `_site` folder, and deploy to S3 along with the rest.  

(Additionally, it might be nice to add a `make` action for local execution, allowing you to do the same from your machine, though this is more of a nice-to-have.)

### There’s unnecessary complexity

This PR combines **content** and **execution** in the source, adding special provisions and front matter. While this approach technically works, there’s a much better solution that is intended for use cases like this one: [project profiles](https://quarto.org/docs/projects/profiles.html).  

With project profiles, you can move all process-related configuration into an environment file and dynamically pull in front matter. See [this workflow](https://github.com/validmind/documentation/blob/5c92efaf696140d56c66b386e75616d102a494b6/.github/workflows/validate-docs-site.yaml#L55) for an example.

This approach should eliminate the need for a separate `_EXECUTED` notebook altogether:  

- **Global execution settings** — Handled separately via a dedicated workflow, not tied to the regular docs site render
- **Front matter** — Pulled dynamically during execution, allowing you to maintain a single-sourced notebook that renders or executes exactly the way you want

With execution decoupled from the regular render, the existing workflow can revert to its simpler, original form. The API key and secret you’ve added should stay, though — they’re valuable for future enhancements.

OK, that's as far as I got after a couple of hours of musing on your PR ... You did some great work here!

validbeck:
> vm.preview_template() — important in the model development process and repeatedly referenced
> vm.tests.run_test() — important to show what ValidMind can do

These both work for me, and on the hosted example. Are you looking somewhere else? (Perhaps WITHIN the training itself, which is the OLD top-level non-executed notebook?) 🤔 

Screenshot 2024-11-19 at 3 58 20 PM

validbeck:
## Adjustment TDL

- [x] vm.preview_template() & vm.tests.run_test() — SOLVED: Nik was looking at the wrong version of the rendered notebook as per the "Developer Fundamentals iframe" note in the original PR comment. These work as expected.
- [ ] `WARNING(validmind.vm_models.test_suite.runner): Config key 'my_test_provider.ConfusionMatrix' does not match a test_id in the template.` — **This should be solved on next execution as Nik added the required test to the model; I will double-check on next run.** For this particular notebook, us executing it via the docs site is meant to be a stop-gap so we agreed this workaround was acceptable for now.
- [ ] Split the execution into its own workflow that triggers only when the notebook changes (See "Execution workflow" below)
- [ ] Remove the _EXECUTED notebook altogether by switching execution to a project profile (See "Project profile" below)

### Execution workflow

#### Part 1

- [ ] Create `.github/workflows/execute-notebooks.yaml` to take care of the live execution of `intro_for_model_developers_EXECUTED.ipynb`
- [ ] Test to make sure it works 
- [ ] Set it to only trigger only when `site/notebooks` (or even just the one notebook) is updated to reduce the regular render runtime*

#### Part 2

- [ ] Add `make` action to Makefile for local execution*
- [ ] [Have the GitHub workflow create the `.env` file on the fly](https://stackoverflow.com/questions/60176044/how-do-i-use-an-env-file-with-github-actions) so we're able to to continue to store the credentials on GH AND get rid of the `_EXECUTED` notebook edgecase*

### Project profile

#### Part 1
- [ ] Remove global execution settings from `_quarto.yml`
- [ ] Create a `_quarto-notebooks.yml` profile so that the front matter from `intro_for_model_developers_EXECUTED.ipynb` can be included and removed from the notebook but still execute this notebook using the new execution workflow

#### Part 2

- [ ] Update the relevant source notebooks in `validmind-library` repo to use the store credentials in `.env` file example already used in the 1st draft to get rid of the `_EXECUTED` notebook edgecase
- [ ] Pull in the latest source from `validmind-library` and test that the `execute-notebooks.yaml` triggers only on notebook updates
- [ ] Confirm that the `.env` file created by GitHub works and the notebook executes properly after the adjustments

### *
I'm an idiot and will likely need help with these. I will fumble around a bit and if I can't figure it out I will holler.

## Other notes

@nrichers Trying to figure out why we metaphorically got our swords crossed yesterday — feel free to tell me off if this is something I should just be working out on my own, or if I crossed any boundaries. 

I have a feeling our tired brains just got stuck on different details rather than actually disagreeing on the ideal outcome? I really didn't mean to make it a big huge thing, or have you feel like you have to walk on eggshells when you give out feedback!

- We agree that live execution of the training notebook needs some refinement, even if the initial solution I proposed is a big leap forward. 
- I want to try implementing your suggestions and it's likely the effort won't be wrapped up in this Sprint whichever way we decide to solve it — this is OK and fine by me, but I don't know if that message was conveyed as maybe me launching into problem-solving mode glossed over that.
- Here is where I think we started to misalign:
     - Doing it all "at once" — i.e. getting rid of the executed edge case now is "easier" because it _is_ technically fewer steps and is cleaner. I concur that this is the most efficient method. (Although, it does rely on being able to edit the source first, which is it's own separate task if we want to do it properly...)
     - However, I don't quite know what I'm doing yet, and I need some time to learn and understand the steps. Breaking down the tasks into 1. Figure out how to execute the edge case as a starting point and set up the infrastructure for future use then 2. Clean up the notebooks so that the source accommodates for this new infrastructure (at which point we can get rid of the edge case) is easier for _me_ as the person who actually has to do the work, because it gives me some breathing room to figure things out. 
     - This also leans into the philosophies of "progress over perfection" and "atomic commits" — making reasonable changes bit by bit instead of dramatic, huge changes. Also, I don't think my efficiency is in question overall, even if I do this in two steps, given our previous conversations about you being amazed by my speed...
- I'm aware you technically have bosshammer privileges — if you really want things done a certain way, you have override powers, so maybe I should just suck it up? :sweat_smile: 

### External dependencies

>  This dependency highlights a limitation with executing a notebook where core functionality relies on external actions outside of the notebook. 

This would be true even if we started hosting executed notebooks on JupyterHub. We'd essentially have to do the same thing: Execute the notebook manually, check to make sure everything ran correctly, have a specific organization and model attached to each notebook, etc. "Automatically" executing notebooks isn't really as simple as pressing "run-all cells." 😬 

Unless, of course, someone smarter than me is able to build a bot that uses the platform in the background whenever a notebook is executed...

validbeck:
re: Testing on https://github.com/validmind/documentation/pull/528/commits/d4e6cfb80cbfb9190ffbe534d922ef1d9ed9a649:

- Workflow ran fine and produced the executed content: https://github.com/validmind/documentation/actions/runs/11942567405/job/33289887695
- **However, since you can't run more than one profile at once this conflicts with the heap tracking.** What's inserted by default is the production info 
- The `page-layout: full` doesn't seem to want to play nice either. Annoying. Not sure why this worked fine in the front matter but hated it on the profile.

As I work on this I realise we may have to actually start splitting up our workflow files and have them lean on each other with some IF conditions instead to minimize the amount of them we have. I'll try to work this out in a Miro board.

validbeck:
Holy forking shirtballs, I figured it out. **Updated Miro workflow:**

New Docs Infra Workflow

Proof that it works conditionally based on what files are included in the PR: **https://github.com/validmind/documentation/pull/536** 

## Part 2

> sc-0000

- [ ] Update the relevant source notebooks in `validmind-library` repo to use the store credentials in `.env` file example already used in the 1st draft to get rid of the `_EXECUTED` notebook edgecase
- [ ] Pull in the latest source from `validmind-library` and test that the updated workflows trigger only on notebook updates (`make get-source`)
- [ ] Have the GitHub workflow create the `.env` file on the fly so we're able to to continue to store the credentials on GH AND get rid of the `_EXECUTED` notebook edgecase
- [ ] Confirm that the `.env` file created by GitHub works and the notebook executes properly after the adjustments

#### nts

- [ ] Retool the current conditional execution steps referencing the execution actions to create the `.env` file on the fly first
- [ ] Only call the execution step if the `.env` file creation isn't skipped 

```yml
- name: Filter changed files
  uses: dorny/paths-filter@v2
  id: filter
  with:
    filters: |
      notebooks:
        - 'site/notebooks/**'

# If yes then create the .env file for use in execution step
- name: Create .env file
  if: steps.filter.outputs.notebooks == 'true'
  id: create_env
  run: |
    echo "APP_ENV=production" >> .env
    echo "PLATFORM_API_HOST=${{ secrets.PLATFORM_API_HOST }}" >> .env
    echo "PLATFORM_API_KEY=${{ secrets.PLATFORM_API_KEY }}" >> .env
    echo "PLATFORM_API_SECRET=${{ secrets.PLATFORM_API_SECRET }}" >> .env
    echo "PLATFORM_DEV_MODEL=${{ secrets.PLATFORM_DEV_MODEL }}" >> .env

- name: Display .env file for debugging, remove on prod
  run: cat .env

# Only execute the demo notebook if .env file creation isn't skipped
- name: Execute demo Intro for Model Developers notebook
  if: ${{ steps.create_env.outcome != 'skipped' }}
  uses: ./.github/actions/demo-notebook
  id: execute-demo-notebook
  with:
    env_file: .env
```

### Nice to have

- [ ] Look into this GitHub workflow environment warning:
```
In addition, using fork() with Python in general is a recipe for mysterious
deadlocks and crashes.

The most likely reason you are seeing this error is because you are using the
multiprocessing module on Linux, which uses fork() by default. This will be
fixed in Python 3.14. Until then, you want to use the "spawn" context instead.

See https://docs.pola.rs/user-guide/misc/multiprocessing/ for details.
```
- [ ] Add `make` action to Makefile for local execution

site/_quarto-exe-demo.yml
```diff
@@ -0,0 +1,12 @@
+format:
```
nrichers, line 1: Maybe not for this PR — because this is already like watching _Gojira_ unleash its full power on the unsuspecting populace — but it might make sense to change all the “demo” stuff to be “development”, to match the Heap stuff which in turn would match what the code does. 

All the demo stuff is madey-uppy and goes back to Andres and I riffing on the prehistoric proto docs site. This also explains why the [PR previews use the `development profile](https://github.com/validmind/documentation/blob/main/.github/workflows/validate-docs-site.yaml#L35). At some point, I hope to phase our all references to "demo".

nrichers (approved):
> Holy forking shirtballs, I figured it out. 

### Holy mackarel, you’re not kidding. What did we unleash by asking you to tackle the switch to profiles, Beck?! 

The single-sourced actions are a great touch. 

I can't even ... Spent about an hour poking around in this PR and I've accepted that we simply need to merge this. (Also, I don't think I have a reliable way to test if this will actually deploy to staging and prod without serious shenanigans, so let's just merge and watch for potential issues.)

site/tests/model_validation/ragas/ContextUtilization.md
```diff
@@ -1,87 +0,0 @@
-# ContextUtilization
```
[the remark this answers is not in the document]
validbeck: Agreed, but I am not sure I am the one who can help solve this scenario. 😆 I assume you are just thinking out loud as a reminder to maybe bring this up with product management? (Since we really only get notified when we retrieve the source and find out something was removed...)

site/_quarto-exe-demo.yml
validbeck: > but it might make sense to change all the “demo” stuff to be “development”, to match the Heap stuff which in turn would match what the code does.

Hm, yes... I did confuse myself a few times doing the Heap bits because of that discrepancy. The problem is our convention for the platform is also... not this, since "development" on that end means production!

validbeck:
@nrichers And staging works!!! 

**Workflow:** https://github.com/validmind/documentation/actions/runs/12021785963/job/33512891350
**Output:** https://docs-staging.validmind.ai/training/developer-fundamentals/developer-fundamentals.html#/section
**Heap tracking:** https://github.com/validmind/documentation/blob/6ee0dcc45ca293ad3e786272fd1bcf59e4b21cf2/site/environments/heap-staging.html#L3

Screenshot 2024-11-25 at 5 24 23 PM

nrichers:
> @nrichers And staging works!!!

Yup, that looks GREAT! I think this makes the fully executed notebook with the complete output the preferred approach to how we surface the stuff in training — it can be the reference point against which people can check their own work. 

### Applause!

validbeck:
@nrichers There seems to be not QUITE success for merging into `prod`: https://github.com/validmind/documentation/actions/runs/12039272835/job/33566716893

It looks like there is some additional logic comparing `main` to `prod` (which makes sense) which means that the filter step returns `false`... hmmmm.

nrichers:
> @nrichers There seems to be not QUITE success for merging into `prod`: https://github.com/validmind/documentation/actions/runs/12039272835/job/33566716893
> 
> It looks like there is some additional logic comparing `main` to `prod` (which makes sense) which means that the filter step returns `false`... hmmmm.

Let's take a look when we meet later today? But looking at that run, the workflow likely needs to compare staging against prod, though even then a `local git` diff I run doesn't seem to pick up the changed notebooks.

PR #539, opened 2024-11-21
Replace Output Template Notebook from Highlighted Code Samples

johnwalz97:

Since we are dropping support for output templates, the code sample notebook for output templates is no longer relevant. This change replaces it with a more relevant notebook on integrating external test providers.


validbeck (approved):
Thanks @johnwalz97! 

Looks like we were already using the external test provider notebook in most other places, so this should be sufficient.

PR #544, opened 2024-11-25
Add login button to docs site

nrichers:

This PR adds a **Login** button & menu to our docs site. If this format works for you, we can ask CSTMR to add the same button to our main site: [**LIVE PREVIEW**](https://docs-demo.vm.validmind.ai/pr_previews/nrichers/sc-7484/add-login-button-to-docs-site/index.html)

image

As part of this change, I'm also switching the search over to the navbar icon — it helps with spacing for the Login button and helps differentiate how search on the front page is subtly different from search that is available elsewhere. 

How it works: 

## External Release Notes

We added the option to log in to the ValidMind Platform to our docs site, making it easy for you to find the right login. Click **Login** to try it out.

### DIFF
--- a/site/_quarto.yml
+++ b/site/_quarto.yml
@@ -74,6 +74,7 @@ website:
     right:
       - text: "Login"
         menu:
+          - text: "Public Internet"
           - text: "{{< var validmind.platform >}} · `US1` {{< fa right-to-bracket >}}"
             href:  https://app.prod.validmind.ai/
             target: _blank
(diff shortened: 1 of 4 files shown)


validbeck (approved):
I like it! (And I see we are *finally* going with the topnav search! 😁) 

The only thing is this weird gap in the staging preview that doesn't show up in the local. Not sure why. 🤔 

Screenshot 2024-11-25 at 4 55 34 PM

I fiddled a bit with the CSS padding on the hero images and it seems to be fine now, but WTF.

nrichers:
> Is there a reason why we decided this is now worth the extra click? I'm in favour of it, just trying to understand the logic of what else shifted.

Yes, we reimplemented the search on the front page of the docs, in preparation for future enhancements like the chatbot. Given how the old search box was implemented on front page ('I macgyvered it'), we could not switch to the search icon without losing the search box.

PR #548, opened 2024-11-26
Update log-in-to-validmind.qmd

mehdi0501:
suggesting deleted the comments about the relative security of public internet vs private endpoints.  Users who know, will know the difference anyways. Others could get the wrong impression that the standard / non-VPV login is not secure, even though it is.


nrichers (approved):
LGTM! 🚀

PR #561, opened 2024-12-03
1 last attempt at the execution workflows

validbeck:

OK, https://github.com/validmind/documentation/pull/559 didn't work — still checking against `main` since that's what the originating commit that is being pushed came from.

I am trying this ONE LAST THING, which is explicitly fetching the `base` branch and calling that instead of relying on the `${{ github.event.before }}` variable. If this doesn't work, then yes, I give up. 

### deploy-docs-staging.yaml

```yaml
- name: Fetch staging branch
  id: fetch_staging
  run: git fetch origin staging

# Checks against the previous commit to staging prior to the current push event
- name: Filter changed files
  if: ${{ steps.fetch_staging.outcome == 'success' }}
  uses: dorny/paths-filter@v2
  id: filter
  with:
    base: origin/staging
    ref: ${{ github.sha }}
    filters: |
      notebooks:
        - 'site/notebooks/**'
```

### deploy-docs-prod.yaml

```yaml
- name: Fetch prod branch
  id: fetch_prod
  run: git fetch origin prod

# Checks against the previous commit to prod prior to the current push event
- name: Filter changed files
  if: ${{ steps.fetch_prod.outcome == 'success' }}
  uses: dorny/paths-filter@v2
  id: filter
  with:
    base: origin/prod
    ref: ${{ github.sha }}
    filters: |
      notebooks:
        - 'site/notebooks/**'
```

### DIFF
--- a/.github/workflows/deploy-docs-prod.yaml
+++ b/.github/workflows/deploy-docs-prod.yaml
@@ -30,13 +30,18 @@ jobs:
             exit 1;
             }
 
+      - name: Fetch prod branch
+        id: fetch_prod
+        run: git fetch origin prod
+
       # See if site/notebooks/ has updates
-      # Checks against the previous commit to the branch prior to the current push event
+      # Checks against the previous commit to prod prior to the current push event
       - name: Filter changed files
+        if: ${{ steps.fetch_prod.outcome == 'success' }}
         uses: dorny/paths-filter@v2
         id: filter
         with:
-          base: ${{ github.event.before }}
+          base: origin/prod
           ref: ${{ github.sha }}
           filters: |
             notebooks:
--- a/.github/workflows/deploy-docs-staging.yaml
+++ b/.github/workflows/deploy-docs-staging.yaml
@@ -30,13 +30,18 @@ jobs:
             exit 1;
             }
 
+      - name: Fetch staging branch
+        id: fetch_staging
+        run: git fetch origin staging
+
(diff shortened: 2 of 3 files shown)


nrichers (approved):
I'd be somewhat surprised if `origin` made a difference since it's typically just some arbitrary remote alias, but LGTM. :shipit: 

FWIW, if this PR doesn't work, I'd recommend simplifying your approach — the amount of effort to create three different execution profiles to be able to track how people use a single notebook that gets embedded in training content we already track is likely overkill.  

https://media2.giphy.com/media/3og0IV5cAmtbkeKPBe/giphy.gif?cid=5a38a5a2whfczzhtgn9cjacw4kiazgq1w7xopoe2ewco6xl6&ep=v1_gifs_search&rid=giphy.gif&ct=g

validbeck:
> FWIW, if this PR doesn't work, I'd recommend simplifying your approach — the amount of effort to create three different execution profiles to be able to track how people use a single notebook that gets embedded in training content we already track is likely overkill.

The issue isn't the execution profiles, FYI — it's just the "filtering" of changed files. The execution runs fine without the filter. 

The only way to simplify this if this doesn't work is to remove the "only execute if `site/notebooks` are changed" on `staging` & `prod` — which I think is an OK compromise, since those run in the background due to the chained workflow pushes (which is why they're SO annoying).

PR #567, opened 2024-12-04
Live notebook execution workflows, take ... whatever

validbeck:

I keep saying I'm giving up but then I don't... anyway. 

The LAST attempt gave me some interesting info in the logs. It did properly check against the correct `base` (`staging`) with the current commit aligning to the PR branch that was being merged in, but it still told me there was no difference in files:

Screenshot 2024-12-03 at 5 15 39 PM

Presumably because when the `merge-main-into-staging.yaml` workflow is complete, there ARE no changes in between `staging` and the commit being compared, because it's already been smushed in...

And so, I am trying to explicitly call the commit PRIOR to the most recent commit:

```yaml
# Checks against the previous commit to staging prior to the current push event
- name: Filter changed files
  if: ${{ steps.fetch_staging.outcome == 'success' }}
  uses: dorny/paths-filter@v2
  id: filter
  with:
    base: staging~1
    ref: ${{ github.sha }}
    filters: |
      notebooks:
        - 'site/notebooks/**'
```

I am hoping this will work. When I call `git show HEAD~1` when I am on staging, I get the following info:

Screenshot 2024-12-04 at 11 21 49 AM

Which aligns with the last commit PRIOR to the last commit that was auto-merged in by the `merge-main-into-staging.yaml` workflow, so in theory the next time the workflows run it will retrieve the right commit to compare against. In theory. (It would help if I knew what I was doing.) 

Screenshot 2024-12-04 at 11 25 19 AM

### DIFF
--- a/.github/workflows/deploy-docs-prod.yaml
+++ b/.github/workflows/deploy-docs-prod.yaml
@@ -18,7 +18,7 @@ jobs:
       - name: Check out repository
         uses: actions/checkout@v2
         with:
-          fetch-depth: 0  # Fetch all branches
+          fetch-depth: 0  # Fetch all history
   
       - name: Fetch Quarto
         uses: ./.github/actions/fetch-quarto
@@ -43,7 +43,7 @@ jobs:
         uses: dorny/paths-filter@v2
         id: filter
         with:
-          base: prod
+          base: prod~1
           ref: ${{ github.sha }}
           filters: |
             notebooks:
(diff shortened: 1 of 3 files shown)


nrichers (approved):
I'm kinda curious what specifying `prod~1` as the base will do. My SWAG is that it will always trigger the workflow — it _should_ — so you might want to make sure the filter for the notebooks works as expected. 

Other than that, LGTM. 🚀

PR #569, opened 2024-12-04
Execution workflows are driving Beck insane

validbeck:

ohmygawddd i hate that you can't just test a workflow locally firstttt

[Turns out it hates the `~1`.](https://github.com/validmind/documentation/actions/runs/12168046796/job/33937973632) The weird thing is, the FIRST execution of the staging filter found changed files, without extra parameters, but I DON'T UNDERSTAND WHY IT WORKED. 💀 [Especially since I just inspected that workflow and it's checking the files against `main`.](https://github.com/validmind/documentation/actions/runs/12021785963/job/33512891350) ????

I just noticed this in the docs though, which is slightly different than what I tried before (previously I tried `${{ github.head_ref }}`):

Screenshot 2024-12-04 at 1 14 19 PM

### DIFF
--- a/.github/workflows/deploy-docs-prod.yaml
+++ b/.github/workflows/deploy-docs-prod.yaml
@@ -43,7 +43,7 @@ jobs:
         uses: dorny/paths-filter@v2
         id: filter
         with:
-          base: prod~1
+          base: ${{ github.ref }}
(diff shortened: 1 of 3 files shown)


nrichers (approved):
<div><br/></div>

PR #582, opened 2024-12-10
Add support chatbot to docs site

nrichers:

This PR adds the support chatbot to our docs site, along with an updated support page.

### Chatbot integration 

image

### Updated support page

- Updated heading
- New section for opening support tickets
- Updated margin note
- (Same chatbot as now appears on all pages we build)

image

## How chatbot integration works

A file that includes the Zendesk script snippet gets embedded in our Quarto profiles. As long as you are building with a profile, the chatbot will be there, on every page of the docs that we build (that's every page except the Python docs but those will be in Quarto soon).

## How to test the chatbot locally

You need to preview or render the docs site with one of the profile that embeds the chatbot, e.g. 

```bash
quarto preview --profile development --port 4444
```


validbeck (approved):
Nice! Thanks for the quick assist Nik, way quicker than me bumbling around testing things first...

nrichers:
Comment from @ValidSean that I will address: 

image

nrichers:
Two additional changes:

1. We didn't have a section that explicitly refers to opening a support ticket anywhere — especially important if you're just searching for related keywords — so I added one: 

image

2. Also, we had a margin note about getting help from a human via email that gets used in a couple of places, including on our main support page. This footnote now links to the new section rather than just calling out one method for getting help:

image

I also checked other references to using our support email and they make sense as-is.

validbeck:
> We didn't have a section that explicitly refers to opening a support ticket anywhere — especially important if you're just searching for related keywords — so I added one:

I mentioned this to Sean on the call (maybe after you left) but we are in agreement that the preference is to send people to the help centre to log a ticket through the form as that enables users to view other tickets from their org if they are logged in.

nrichers:
> I mentioned this to Sean on the call (maybe after you left) but we are in agreement that the preference is to first send people to the help centre to log a ticket through the form as that enables users to view other tickets from their org if they are logged in.

I added the Help Center option, as suggested and also added a callout to say that this is the recommended option. 

image

validbeck:
Nice! I just cleaned up the styling/wording a bit:

Screenshot 2024-12-11 at 9 35 13 AM

I tried using a preview tile, but Zendesk doesn't allow iframe embeds...

Screenshot 2024-12-11 at 9 35 44 AM

nrichers:
> Nice! I just cleaned up the styling/wording a bit:

I like the general idea of this! It probably falls into a pattern, but this change also introduced just a bit too many visual artifacts, so I did some cleanup (and fixed the working for phone calls).

**Before** 

image

**After**

image

validbeck:
For email, they need to SEND from the address otherwise it won't auto-associate it with the right org in Zendesk — that's why I worded it that way.

validbeck:
Also, I don't like the new layout. Harder to read for me.

What about:

nrichers:
> What about:

Looks good — merged so I can move on with my life. 😬

PR #607, opened 2024-12-23
Edited release script & notebook

validbeck:
## Internal Notes for Reviewers

> sc-7068

### `generate_release_objects.py`

> **TEST OUTPUT**: https://github.com/validmind/documentation/pull/608

- Cleaned this script up
- The `main` function now performs all the same features as the notebook (for example, `auto_summary` was previously missing) 
- Instead of repeating the code in `main` we simply call the granular functions defined earlier in the script — much cleaner, and also if we need to edit any of the functions now we don't have to do it 2-3 times

| Old | New |
|---|---|
|Screenshot 2024-12-23 at 11 11 38 AMScreenshot 2024-12-23 at 11 11 58 AM Screenshot 2024-12-23 at 11 12 09 AM Screenshot 2024-12-23 at 11 12 17 AMScreenshot 2024-12-23 at 11 12 20 AM| Screenshot 2024-12-23 at 12 08 19 PMScreenshot 2024-12-23 at 12 08 27 PM Screenshot 2024-12-23 at 12 08 34 PM |

### `generate-release-notes.ipynb`

> **TEST OUTPUT**: https://github.com/validmind/documentation/pull/608

- Edited this ENTIRE notebook for coherency & adherence to our style guide
- Redid the ToC for updated sections
- Instead of repeating the code in the notebook and importing functions from `generate_release_objects.py` in piecemeal, we simply import the whole script with an alias — "single-sourcing" at its finest 

| Old | New |
|---|---|
| Screenshot 2024-12-23 at 11 18 39 AM| Screenshot 2024-12-23 at 11 18 46 AM|
|Screenshot 2024-12-23 at 11 23 04 AM |Screenshot 2024-12-23 at 11 23 27 AM |

#### Set up OpenAI API

- Updated the "Set up OpenAI API" to ask for the location of your `.env` file (default is set to `../.env`)
- Output confirms the location of the file and whether or not the key is found

**YOU CAN TEST THIS BY INPUTTING A FILEPATH WITH NO VALID .ENV OR REMOVE THE EXPECTED KEY FROM YOUR .ENV FILE**

| Old | New |
|---|---|
|Screenshot 2024-12-23 at 11 36 18 AM | Screenshot 2024-12-23 at 11 36 32 AM|

| Prompt | Output |
|---|---|
|Screenshot 2024-12-23 at 11 24 13 AM |Screenshot 2024-12-23 at 11 24 34 AM |

| No `.env` file at location | No valid OpenAI API key |
|---|---|
[...]

PR #610, opened 2024-12-27
Cleanup of 2024 releases + new cleanup notebook/script

validbeck:

> sc-7885

I collated our 2024 releases with the magic of the new "Year end cleanup" notebook/script: **[LIVE PREVIEW](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-7885/cleanup-of-2024-releases/releases/2024/2024-releases.html)**

### Year end cleanup

> **VIEW DEMO:** https://github.com/validmind/documentation/pull/611

Riffing on the stuff I learned in https://github.com/validmind/documentation/pull/607, now there is a new comprehensive script/notebook that takes care of the following for us:

- Creates a new year directory for the year you specify
- Moves that year’s releases into that directory
- Creates a new `{year}-releases.qmd` page that displays the moved releases as listing tiles
- Updates our `_quarto.yml` sidebar with a new accordion folder for that year and
    - Move the `# CURRENT-YEAR-END-MARKER` to above your rounded up releases for that year
- Resolves any broken absolute links to the moved release pages
- Returns a list of relative filepaths within the moved release pages that may need to be adjusted
- Shows you the list of files to be committed, omitting the changes in the `release-scripts` folder
- Runs a preview using Quarto

#### Notes

*Whether you run the notebook or the script, the functionality is the saaaaaaaaame!* Amazing.

- Smart de-duping. Only creates the yearly directory or yearly listing page if they don't already exist.
 - Listings are only added to the yearly page  if it was previously empty; no duplication if you accidentally run the script for a year that already has been rounded up.
- Only moves release directories it can find for that year.
- Sidebar marker move inserts it right above the collated year, so for example if we already have a release for a year ahead it won't put it above that and mix up the automation/our process.

#### Updated internal process documentation

Wow, the old process was so tedious. I am so glad I figured this out! **[I updated our internal docs to match](https://www.notion.so/validmind/Create-Release-Notes-c4291bb92f644c5eb490cc21a8a79ca5?pvs=4#0fd19127eb128090b9fde4965384be5a):**

| OLD | NEW |
|---|---|
| Screenshot 2024-12-27 at 3 13 32 PM Screenshot 2024-12-27 at 3 13 53 PMScreenshot 2024-12-27 at 3 14 00 PMScreenshot 2024-12-27 at 3 14 05 PM| Screenshot 2024-12-27 at 3 35 19 PM|

### Generate release notes

I improved this further, as I still didn't think it was as polished as it could have been.

#### Output refinement

Outputs are now... helpful, descriptive, and more human readable, whether you run this in the notebook or via the Makefile. Yay! 

For example... `No PR summary found` vs `No PR summary found for #1095 from validmind/frontend` etc.

| Old notebook output | New notebook output |
|---|---|
| [View old demo run](https://github.com/validmind/documentation/blob/024037e1fcedbbc1fa0c4f84724b11e4222202d0/release-scripts/demo-runs/generate-release-notes_OLD.ipynb) | [**View new demo run**](https://github.com/validmind/documentation/blob/024037e1fcedbbc1fa0c4f84724b11e4222202d0/release-scripts/demo-runs/generate-release-notes_NEW.ipynb) |

#### Collect GitHub URLs

- If you insert a duplicate release tag URL (which I have absolutely done) it will dedupe that instead of including it twice in the generation!
- Counts the tags for each repo for you, just so you can check if you've entered everything you wanted to easily.

Screenshot 2024-12-27 at 3 50 19 PM

#### Create release folder and file

I realised that the old version just... rewrote the output created if you entered the same date again. Which is fine, if that's what you wanted (for example, if you wanted to regenerate everything with ChatGPT, or made a mistake in entering your URLs)... but... just in case, since I love thinking about edge cases, now there's a prompt that asks you if you're sure if that scenario occurs:

| Prompt | If you say "no" |
|---|---|
|Screenshot 2024-12-27 at 2 38 05 PM | Screenshot 2024-12-27 at 2 38 18 PM|

#### Update sidebar

Again, if you entered the same date again it would simply add a new entry to `_quarto.yml` resulting in duplication of links in the sidebar. Now in the situation of that edge case, it doesn't. 

#### Update index

Same thing here — if you entered the same date again there was an edge case. While it wouldn't add a duplicate entry it would keep removing the oldest entry even if a latest release entry wasn't added. Now it doesn't do that.

#### Show files to commit

Why were we using a Python script to do this previously? I have no idea. I was fighting with it for hours to do some weird filtering shenanigans then I just realised... huh. We can simplify it. Now it just calls `git status` and BAM, doesn't show the changes in `release-scripts`. 

| Old | New |
|---|---|
|Screenshot 2024-12-27 at 3 59 58 PM |Screenshot 2024-12-27 at 3 59 47 PM |


nrichers (commented):
I know this is already merged but LGTM! This PR includes some very useful changes. 🥳 

> Resolves any broken absolute links to the moved release pages

We might want to consider NOT moving release notes after the fact. This would mean a release in 2025 would always live in a `2025` path and not get relocated at the end of the year. This approach aligns with the general principle that releases should be static—once published, they shouldn’t be added to, updated, or moved.

**Why this matters:** Release notes are often linked from external sources. For example, I had a blog post in progress that referred to our latest 2024 release notes via a direct URL I tested in December. During a final sanity check today, I noticed the link was broken because the release notes had been moved in the meantime. 

Alternatively, we could use aliases, but since we already organize content extensively with folders, sticking to folders seems simpler. (Also, I’ve been burned by aliases/redirects more than once, and I predict we’ll encounter similar issues when we least suspect it.)

validbeck:
@nrichers Thank you for taking the time to review; as discussed I agree and I have created a Story for the follow-up: https://app.shortcut.com/validmind/story/8017/release-notes-scripts-notebooks-qol-improvements

PR #624, opened 2025-01-10
Fixed anchor placement for Jupyter ToC extension

validbeck:

> sc-4131

### validmind-toc

Screenshot 2025-01-10 at 2 18 34 PM

[**Refer to the README for full details**](https://github.com/validmind/documentation/blob/0f186c1528596303cf8397233f6870b61540e5c7/validmind-toc/README.md) — Includes instructions on how to update the extension mostly for my own reference because I know I will forget if I have to do it again

I forked the extension we've been using and created a modified version of it. This fixes:

- Page anchors being inserted after the markdown header instead of above it in its own line, requiring us to re-open the notebook in text view and using regex to find & replace all the injected ToC anchors
- Being unable to update or remove ToCs after we modify them with the extension, requiring SO much #&*ing manual editing any time we made changes to a notebook

#### Try it yourself

1. - [ ] Grab this PR: `gh pr checkout 624`
2. - [ ] Follow the instructions on the [**User guide**](https://github.com/validmind/documentation/blob/0f186c1528596303cf8397233f6870b61540e5c7/validmind-toc/installation/README.md)

Bam, add and remove the ToC to our specs on a random notebook that doesn't already have a ToC of your choice, edit the notebook, update the ToC, whoo!!!


nrichers (changes requested):
I've reviewed this pull request and tested it locally and it generally seems to do what it says on the tin, which is great. 🎉🎉🎉 

That said, I don't think this extension for VS Code belongs into our repository and I have some concerns about building something on top of the work of others without contributing back to the open source community. There are also some licensing issues this PR would create which are inherently problematic.

Let me explain ...

## Building on the work of others

This pull request builds on the work done by xelad0m and made public as the [Jupyter TOC](https://marketplace.visualstudio.com/items?itemName=ms-toolsai.jupyter) extension in the Visual Studio Marketplace.

From an authorship perspective, it feels problematic to build on this code base and then pass it off as a "ValidMind TOC" extension that is not made generally available to others unless they happen to discover it in our repository — licensing considerations aside, it feels like the wrong thing to do.

A much better way would be to update the original extension or to create a public fork.

## Licensing considerations

Our documentation repo is licensed under the AGPL-3.0 license which is a strong copyleft license. The existing Jupiter TOC extension is a licensed under the Apache license. 

Combining code under these licenses is problematic because the licenses are not inherently compatible (note that AGPL-3.0 != GPLv3 which IS compatible with Apache 2.0):

- AGPL-3.0 requires software incorporating its code to also be licensed under AGPL-3.0 when distributed; its strong copyleft overrides other licenses, forcing combined code to adopt AGPL-3.0 terms.
- The Apache License is permissive, allowing use in projects with various licenses, including proprietary ones.
- Apache includes a patent grant clause that terminates if the licensee files a patent lawsuit against contributors.
- AGPL-3.0 lacks this patent clause and cannot enforce similar terms.
- AGPL-3.0’s strict copyleft is incompatible with Apache’s permissiveness, which does not require derivative works to use the same license.

The only way around this would be to dual license different parts of our documentation repo if we include this code but this is not trivial, either — it took us months to sort out our dual AGPL-3.0 and commercial licensing. 

The best answer is to not introduce code that has incompatible licensing into our documentation repo at all.

## Where do we go from here?

Above all, we should strive to be the best citizens that we can be in the open source world. This approach makes good sense since we ourselves have open source  and we want our authorship to be respected.

My recommendations, in order of preference:

- Open a pull request to make these changes in https://github.com/xelad0m/vscode-jupyter-toc
- Fork https://github.com/xelad0m/vscode-jupyter-toc and make the changes there to create a new extension that anyone can install.

I would also recommend avoiding TOC option names that include `validmind`, as these options don’t appear inherently unique to our use case and could work with other Jupyter Notebooks.

validbeck:
After some discussion with @nrichers we have decided to put off reimplementing ToC in our notebooks. While the modified extension does what I intended it to do, it's simply a band-aid on top of the actual problem we have with the notebooks.

- Currently, our notebooks are a bit gigantic, hence necessitating the embedded table of contents within the notebook itself. This is mostly to accommodate the fact that these notebooks are currently meant to be opened in our JupyterHub demo environment (which may change in the future). 
- However, since we render the same notebook files as `.html` pages on our documentation site, the original table of contents extension conflicts with how the table of contents is constructed natively in Quarto (as it already has support for ToC/options for in-page ToCs), necessitating our annoying find-replace workaround that essentially disables the functionality of the extension for that notebook after the first application of the table. 

The *better* solution is to reevaluate how we structure our monolithic notebooks and reimplement how we present them as we audit them ([Epic-7898](https://app.shortcut.com/validmind/epic/7898?cf_workflow=500000037&ct_workflow=all&group_by=owner_ids&vc_group_by=day)) making them individually shorter and easier to consume potentially without strange workarounds/additional extensions.

I have moved the updated extension functionality to its own repo here for future reference/for now (I also added a neat setting that allows us to toggle off the embedded ToC for HTML pages rendered by Quarto if we do want to use this stopgap in the meantime as it bothered me that it duplicated it for the docs site — checked the render and it looks great): https://github.com/validbeck/jupyter-notebook-toc

 | Settings | ToC Cell |
|---|---|
|new-settings | content-hidden-html content-hidden-html-cell|

Screenshot 2025-01-15 at 1 41 54 PM

nrichers (commented):
> I have moved the updated extension functionality to its own repo here for future reference/for now

@validbeck thank you for making the changes we discussed — I took a quick look at the forked TOC extension — the ability to hide the ToC when the output is HTML is a nice touch. 

What's left in this PR for me to review/approve? With the extension in a separate repo, I think this PR can be closed.

validbeck:
> What's left in this PR for me to review/approve? With the extension in a separate repo, I think this PR can be closed.

Yep, just wanted you to see the comment before I closed it! All good now.

PR #638, opened 2025-01-30
[DEMO] New site mockup

validbeck:

> sc-7863

Combining the sum of our ideas for the presentation. Obviously not a final draft, just to demonstrate the direction we want to move towards. 

[**LIVE PREVIEW**](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-7863/solidify-ideal-new-site-structure/index.html) | [Demo slides](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-7863/solidify-ideal-new-site-structure/training/site-mockup/site-mockup.html#/title-slide)


site/developer/get-started-validmind-library.qmd
```diff
@@ -37,44 +38,73 @@ listing:
...
+{{< video https://youtu.be/rIR8Mql7eGs title='ValidMind QuickStart' >}}
+
+:::
+
+::::
 
-The {{< var validmind.developer >}} provides a rich collection of documentation tools and test suites, from documenting descriptions of your dataset to validation testing your models for weak spots and overfit areas. 
- 
-<!-- Using the variable in alt text messes up the image display  --> 
+## Develop models with ValidMind
```
nrichers, line 72:
```suggestion
## Document and test models with ValidMind
```

It's true that you are developing models as you are documenting and testing, but the thingamajig that ValidMind adds is the latter.

nrichers (commented):
Assorted initial impressions, thank you for putting together this PR! 🍰🍰🍰

**Navbar**

image

image

The more I see of them, the more I like the clean look and feel of your navbars in the developer and training portals. I think we should just figure out a similar design for the home page and user guides ... 

**Hero banner**

image

This banner feels like it should be half the height. On a small laptop screen you don't see enough of the getting started content tiles and the banner itself doesn't add much. 

**Content tiles**

Pasted Graphic 2

We can tinker with the content of these tiles but the general feeling should be one of _simplicity_:

- Use the full screen width to give more space to the content
- Likely, this means loosing the background
- Do include **Get started** buttons for a clean, simple CTA 
- The developer tile needs to show `import validmind as vm` code￼ 
- "Setup + Administration" should likely just be "Onboarding" or something simple like that, all we're covering here is the initial getting started tasks

**Platform value proposition**

Pasted Graphic 3

I like where this is going! That said, let's not get too much into the weeds with the graphics and design —I asked Rod for some of Jeffrey’s time today after talking to Mehdi. 

**Training content**

image

This part of the front page is largely untouched, and it feels that way. To me, it’s the least well-defined section, and I’m not convinced we even need it here. If we do, it should focus on our training program more generally rather than embedding specific courses. I’d be fine with removing it for now, which would make reaching our release notes faster.

**QuickStart pages**

image

You were right about having dedicated pages for these. I like the simplicity here and the freedom it gives us to embed this content elsewhere. 

## Nice work! 

Let's just leave this PR up for now as we iterate over ideas and the design? There's much more to come, I'm sure.

validbeck:
> The more I see of them, the more I like the clean look and feel of your navbars in the developer and training portals. I think we should just figure out a similar design for the home page and user guides ...

I think we should keep it on the homepage, but agree that guides can be squashed down when we get to retooling guides. Personally, I hate it when homepages make you DIG for the sections. Sometimes as a user, I have a goal, and I know exactly what I'm looking for; don't divert me from that by removing choices for navigation. 

> This banner feels like it should be half the height. On a small laptop screen you don't see enough of the getting started content tiles and the banner itself doesn't add much.

I agree, but again — I tried to keep this light and not "me redesigning everything as that's essentially just doing the work instead of drafting." Here's an adjustment though:

Screenshot 2025-02-05 at 11 56 11 AM

> Use the full screen width to give more space to the content

I disagree here; negative space is important in both design and readability. 

> Do include Get started buttons for a clean, simple CTA

I didn't like that all of the buttons had the same text and prefer the listing tiles. Again "nebulous" links are confusing for users. I agree that the image tiles aren't necessary, I was just playing around with the looks. I'm OK with this compromise:

Screenshot 2025-02-05 at 11 57 35 AM

> The developer tile needs to show import validmind as vm code￼

Why? What does this add? This is just going way too granular without the proper introduction. To me, that's stuff that should go in the actual guide rather than on the home page.

> "Setup + Administration" should likely just be "Onboarding" or something simple like that, all we're covering here is the initial getting started tasks

Fair enough, no real opinion here, but I will wait on this one until others weight in as it seems to be in between rather than strictly being onboarding.

> I like where this is going! That said, let's not get too much into the weeds with the graphics and design —I asked Rod for some of Jeffrey’s time today after talking to Mehdi.

Yep, just something I slapped together in a couple minutes so it wasn't just inconsistent with the rest of the page having some sort of CSS polish. 

> This part of the front page is largely untouched, and it feels that way. To me, it’s the least well-defined section, and I’m not convinced we even need it here. If we do, it should focus on our training program more generally rather than embedding specific courses. I’d be fine with removing it for now, which would make reaching our release notes faster.

Yep, I agree, but again I think this is out of the scope of our "Get Started" restructure. Guides, Developer portal, and training etc. are all additional initiatives. I think having training on the front page is still important though, but we don't necessarily need all the courses on there. Here's an updated mockup: 

Screenshot 2025-02-05 at 12 24 38 PM

(Again, just a mockup for what content we potentially want to include, not a solid design :p) 

> You were right about having dedicated pages for these. I like the simplicity here and the freedom it gives us to embed this content elsewhere.

If you look at the drop-down menus, there's also the landing page for these still if people want to reach them from NOT the homepage, or to give out to their other users, etc.: 

| Drop-down | Landing |
|---|---|
| Screenshot 2025-02-05 at 12 02 07 PM| Screenshot 2025-02-05 at 12 03 12 PM |

We can quibble about the inclusion of appropriate images later. **Reminder:** Just a mockup so that things don't look empty.

site/training/site-mockup/site-mockup.qmd
```diff
@@ -0,0 +1,276 @@
...
+
+::::
+
+::: {.f2}
+<br>
+Experience a mockup **live** on the next page. {{< fa hand-point-right >}}
+
+:::
+
+## {background-iframe="https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-7863/solidify-ideal-new-site-structure/index.html" background-interactive="true" data-preload="yes"}
```
nrichers, line 143: This slide continues to make me uneasy — after sitting on it, I think the problem might be that _the representation obscures the ideas_. 

- **At a distance:** It's simultaneously too close to the current site to feel like a mockup and too far away from our future site to feel like a realistic mockup. 

- **Up close:** I get mired in the details, like the graphics and the the fact that the tiles aren't as simple as I'd like them to be, and so forth. 

Flipping back and forth between slides 11 - 13, it also feels like there is one slide too many.

One option here might be not to show the home page in its complete form but to _show only excerpts_  — hide the navbar, hide stuff after the three tiles — and combine a variation of slides 11 - 13 into two slides. 

Maybe something to chat about more?

site/training/site-mockup/site-mockup.qmd
```diff
@@ -0,0 +1,276 @@
...
+::: {.tc}
+**Improved support experience**
+:::
+
+::: {.f4}
+As part of flattening our navigation, we can also improve the support experience by combining our main “Support” landing page with our “FAQ” section.
+:::
+::::
+
+## {background-iframe="https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-7863/solidify-ideal-new-site-structure/developer/get-started-validmind-library.html" background-interactive="yes" data-preload="yes"}
```
nrichers, line 208: Generally, I'd like to suggest that the redesign should give a feeling of simplicity or elegance that our current version lacks, even in an incomplete mockup. 

I look at this slide and I again get that feeling that what's represented here covers up the good ideas it's trying to convey. I'm not entirely surprised, as it's basically what we currently have with some tweaks applied.

My thoughts on how to , all up for discussion: 

- I would not show the full page in its complete form but to show only excerpts — at a minimum hide the sidebar and omit the video, it's obsolete under the new QuickStart. 
- We should have a chat about whether landing pages should be more like a home page, without a sidebar. There are too many navigational elements competing with each other. 
- (I think you independently had the same idea, as slide 17 already shows a "Guides" page without a sidebar.)

nrichers (commented):
I left some initial comments — more to come, but I focused on the big-picture ideas while not completely forgetting how mockups can come across for an audience.

site/training/site-mockup/site-mockup.qmd
```diff
@@ -0,0 +1,276 @@
...
+<br>Experience a mockup **live** on the next page. {{< fa hand-point-right >}}
+:::
+:::
+
+::: {.column width="70%" .bl .pl4 .f3}
+- **A clear value proposition**—what ValidMind helps them achieve.
+- **A simple "getting started" path**—highlighting key actions upfront.
+- **A single, focused call to action**—eliminating confusion on what to do next.
+
+::: {.callout}
```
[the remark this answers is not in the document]
validbeck: Yes, that was deliberate — it's meant to be an embedded callout and not an overlay but you're correct that it doesn't seem quite right. 😅 

Overlay looks like this, which also isn't quite right: 

Screenshot 2025-02-12 at 3 43 46 PM

Embed looks like this, which also isn't quite right:

Screenshot 2025-02-12 at 3 48 12 PM

But maybe with some padding? 🎉 

Screenshot 2025-02-12 at 4 15 37 PM

site/training/site-mockup/site-mockup.qmd
```diff
@@ -0,0 +1,276 @@
...
+  data-background-image: "../assets/home-hero.svg"
+skip_preview: true
+---
+
+# Objectives
+
+- **Explain the approach** — We want to streamline user journeys, starting with our getting started experience. Let's talk about how we're going to accomplish that.
+- **Gather team input** — The new documentation site should be a *team* effort; we want your ideas on updated content and design*!
+
+:::: {.tc .mt5 .f3 .embed}
```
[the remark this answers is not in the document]
validbeck: It hurts me to look at and feels like overkill, but I increased the disclaimer text size and changed the colour of the embed:

Screenshot 2025-02-12 at 4 22 44 PM

I even added a watermark to the embedded demo pages:

Screenshot 2025-02-12 at 4 37 40 PM

If anyone during the demo asks me why the messaging around this is so aggressive, you're going to have to field that question. :p

site/training/site-mockup/site-mockup.qmd
validbeck: OK, I added the watermark which should at least solve some of the issue, yes?

I get what you're saying, but this demo ISN'T JUST STATIC. I will be using the slides as a *starting point* as I click around while talking and show people the things we're highlighting. I tried a version where I added some anchors and took away the interactivity of clicking around, BUT the way RevealJS plays on different screen dimensions made that really hard to accomplish. 

> I get mired in the details, like the graphics and the the fact that the tiles aren't as simple as I'd like them to be, and so forth.

I get it on some level, but I think this might also just be a preference situation? What I mean by that is it was way quicker for me to to just adjust some lines of code to get a fake version of the site than it was for me to spend hours making images in Miro to show off, and I feel like our working styles here are conflicting. Again, I could go in and leave a bunch of comments on your Notion document about what I think could be refined, but I don't really think that's a productive approach to take at this juncture. 

**I also want to add:** I think I should get a say in the design too, yeah? YOU want more simple tiles, *I* like some fancier tiles — Why can't *my* version of the proposal reflect some of my preferences? 

Can we compromise at this stage with the two versions and just let the masses weigh in on what they like more? You want us to align at this stage and I think that's not necessary to start the conversation. 

> Flipping back and forth between slides 11 - 13, it also feels like there is one slide too many.

I'll take a closer look at this tomorrow, I agree that there can be some stuff more compressed.

site/training/site-mockup/site-mockup.qmd
validbeck: I think the sidebar should stay for now for the purposes of this demo, but I will go ahead and remove the video as I don't have any particular attachment to it.

I agree that it should look more like the other portals eventually, but I don't really want to spend any more time retooling the contents of that page to encapsulate everything just for the demo. That is 100% "implementation" work at this point, and I have already spent more time than I'd like tweaking this when it was just meant to be a quick draft.

site/training/site-mockup/site-mockup.qmd
nrichers: > Can we compromise at this stage with the two versions and just let the masses weigh in on what they like more? You want us to align at this stage and I think that's not necessary to start the conversation.

I think two versions are fine at this stage, but keep in mind that all the versions are collaborative — there's stuff from my Notion doc in your presentation and vice-versa. Eventually, we'll need something we can demo at a weekly that speaks in a **unified** voice. 

What I would prefer to avoid is three versions (our two plus a third from Design) floating around, each being showcased as the single source of truth. 

How we manage communication with the rest of our team is important.

nrichers:
> OK, I added the watermark which should at least solve some of the issue, yes?

The watermark definitely helps!

validbeck:
Closing this PR demo as most of it has been captured in follow-up initiatives. Just saving these onboarding tiles here just in case we want to return to this version of the text:

Screenshot 2025-06-05 at 9 47 22 AM

PR #640, opened 2025-01-30
Switch Python API reference to Quarto

nrichers:
## Internal notes for reviewers  

This PR replaces the pdoc-generated Python HTML docs with Quarto Markdown source, which is rendered alongside the rest of the site: [**Live preview**](https://docs-demo.vm.validmind.ai/pr_previews/nrichers/sc-4660/python-api-with-quarto/validmind/validmind.html)  

Relates to: https://github.com/validmind/validmind-library/pull/306  

| Before | After |
|--------|-------|
| image | image |

## Modified files  

- `site/Makefile` — Updated `make python-docs` to handle the new Quarto docs  
- `site/_quarto.yml` — Updated to use the new QMD source and removed the post-render action for Python docs  
- `python-docs.zip` (DELETED) — No longer needed with the new Quarto source  

**Bonus:** `make clone` now accepts a branch for PR testing.  

## New files and folders  

Do not review here, post comments to  [#306]( validmind/validmind-library#306:

- `site/validmind/README.md` — How the Quarto Markdown gets generated  
- `site/validmind/_metadata.yml` — Configuration for Python API content  
- `site/validmind/_sidebar.yml` — Generated sidebar for Python docs  
- `site/validmind/validmind.css` — Styling for the API docs  
- `site/validmind/validmind.qmd`, `site/validmind/validmind/**` — Generated API documentation matching the codebase and pdoc structure

### `site/validmind/`

Do not review the Quarto Markdown files here. If you see issues, review them in the PR that generates them: https://github.com/validmind/validmind-library/pull/306

The Quarto Markdown files are a drop-in replacements with only minor differences.

### `quarto preview`  

**Important:** If you test this PR locally, delete the `site/_site folder` first to remove outdated Python HTML files. 

(This step is unnecessary when running `quarto render`, which removes old HTML output automatically.)

### `Makefile`  

To test how the Quarto Markdown source files get from the `validmind-library` repo into this one:  

```sh
gh pr checkout 640
make clean
make clone BRANCH=nrichers/sc-4660/python-api-with-quarto
make python-docs
```

You should see the `site/validmind/` folder disappearing and reappearing with the new source files.

## External release notes  

This release introduces a completely redesigned Python API reference using Quarto instead of the previous HTML-generated docs. The new ValidMind Library documentation features improved navigation with a familiar sidebar, enhanced code signature styling, and integration with our main docs site search.

The documentation structure mirrors the Python package layout, ensuring backward compatibility with our old Python API reference while providing a more maintainable and reader-friendly experience.  

[ValidMind Library](/validmind/index.html)


validbeck (changes requested):
## Makefile

I had to call `cd site` before the commands, but this was expected — your C&P code doesn't work for other reviewers, though. 😅 Otherwise, works like a charm.

## Other comments

Hmmmm. Something is still amiss. I checked the files AND the variables and they don't have any leading `../` but this appears in multiple places whenever there is "ValidMind" happening:

Screenshot 2025-03-03 at 1 58 26 PM

Happening on any page that uses the `{{< var vm.product >}}` variable... 🤔 ... and I mean ANY ...

Screenshot 2025-03-03 at 1 59 51 PM

Seems to be the only variable affected, though. Not sure what's happening at all else I'd just apply a fix. 😬

validbeck:
Small distinction but I think important (as we should probably rename the Developer section to just `ValidMind Library` once we clean up our getting started experience:

Screenshot 2025-03-03 at 2 17 35 PM
Screenshot 2025-03-03 at 2 17 55 PM

validbeck:
The new docs should be 1:1, correct? Then we should ideally update these 216 instances of static `.html` links to the `.qmd` links in the rest of the docs so that Quarto knows to check them on render:

Screenshot 2025-03-03 at 2 21 57 PM

nrichers:
> The new docs should be 1:1, correct?

Thanks for pushing some changes — yes, the new docs should be 1:1 more or less in the HTML output. Check "Output file comparison" in https://github.com/validmind/validmind-library/pull/306 where I list the files side-by side.

nrichers:
@validbeck I found a workaround via 8909b19415dc7f040518f26f6f7dbbbafb182d66: prefixing a zero-width whitespace character (`&#8203;ValidMind`) avoids the issue:

Capto_ 2025-03-04_08-54-48_am

My hypothesis is that somewhere in the metadata for the API reference changes, `validmind` is being interpreted differently. I plan to do a bit more investigation but I if this workaround holds, we might just have to live with it for now.

validbeck:
> @validbeck I'm honing in on the ../ issue: it's the value of the variable, nothing to do with nesting or the variable name itself. I tested with a variety of different variables styles and values and basically anytime the value is set to ValidMind, we get the ../.

So weird it's just happening now — this variable has been like this since we introduced the nesting, for example, and not rendered like this before. 🤔 It's not happening on the prod version of the site, for example, so I wonder if this is related to a recent Quarto update? Seems like they just released on Feb 24, which is definitely after I've published to prod...

validbeck:
> Thanks for pushing some changes — yes, the new docs should be 1:1 more or less in the HTML output. Check "Output file comparison" in https://github.com/validmind/validmind-library/pull/306 where I list the files side-by side.

I haven't updated the `.html` links to `.qmd` yet, let me know if you would prefer I handle that or if you got it!

nrichers:
> I haven't updated the `.html` links to `.qmd` yet, let me know if you would prefer I handle that or if you got it!

I'll handle, please.

site/releases/2025/2025-jan-31/release-notes.qmd
```diff
@@ -374,7 +374,7 @@ We added raw data storage across all {{< var validmind.developer >}} tests. Ever
 :::
 
 ::: {.w-20-ns .tr}
-[`RawData`](/validmind/validmind.html#RawData){.button target="_blank"}
+[`RawData`](/validmind/validmind.qmd#RawData){.button target="_blank"}
```
nrichers, line 377: @nrichers like the original, we currently include a `Class` prefix for classes like RawData but unlike the original that prefix is part of the anchor:

- Investigate if we can modify the Jinja2 templates to include the muted `Class` text via some CSS trickery, which would make anchors work as-is
- Falback plan: Take note of where these links occurs and open a story to fix in the future

nrichers: This should be resolved via a number of commits that remove the class prefix from headings, replacing it with a new `::before` CSS class. 

This change fixes most anchor links but I also noticed that we sometimes use mixed-case anchors in links because the old HTML reference provided them. Quarto generates lowercase anchors as recommended  — I manually fixed those anchor links in our QMD source and a few other, smaller things I noticed (casing, links that were wrong even for the old HTML reference, and in one case we linked to a method to appears to have been removed from the codebase).

validbeck (changes requested):
Since we did have to go with the zero-width whitespace as a workaround, can we please leave a comment on the variable file for future reference? 🙏🏻 (I would do it, but I think I'm still not understanding the root cause of the issue as I wasn't the one that investigated it...)

Otherwise, there was a broken link when we changed the `.html` to `.qmd` that I fixed — we linked to the full docs site URL instead of the relative root `/validmind` folder so that was expected.

site/developer/model-testing/testing-overview.qmd
```diff
@@ -152,4 +152,4 @@ Absolutely! {{< var vm.product >}} supports custom tests that you develop yourse
 
 ## {{< var validmind.api >}} reference
 
-[{{< var validmind.developer >}} Reference](https://docs.validmind.ai/validmind/validmind.html){target="_blank" .button .button-green}
+[{{< var validmind.developer >}} Reference](/validmind/validmind.qmd){target="_blank" .button .button-green}
```
validbeck, line 155:
```suggestion
[{{< var validmind.developer >}} API Reference](/validmind/validmind.qmd){target="_blank" .button .button-green}
```

validbeck (approved):
One last suggestion (my bad for missing it the 1st few rounds) but I think we're ready! 🚀

PR #672, opened 2025-03-13
feat: example implementation of trigger in other repo

nibalizer:

Just an example of how we can trigger a workflow in another repository via repository dispatch. The purpose of this PR is to test triggering another workflow in the `installation` repo to deploy an instance of our docs site with additional content.

### DIFF
--- a/.github/workflows/test_installation_trigger.yaml
+++ b/.github/workflows/test_installation_trigger.yaml
@@ -1,4 +1,4 @@
-name: Trigger job in Installation
+name: Trigger job in the installation repo
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


.github/workflows/test_installation_trigger.yaml
```diff
@@ -0,0 +1,15 @@
+name: Trigger job in Installation
```
nrichers, line 1:
```suggestion
name: Trigger job in the installation repo
```

nrichers (approved):
LGTM! Updated the root comment with some info to explain what the PR does and left a comment with suggested changes you can commit. 🚀

PR #677, opened 2025-03-20
feat: extend build to include private section

nibalizer:

This adds an external dependency for additional private documentation. The gha and Makefiles are setup to track the `main` branch of the dependency repository.

ValidMind Employees can access this link with further explanation of why: https://www.notion.so/validmind/Docs-2-0-Site-18a19127eb1280ada461c96175d449a1?pvs=4#1c019127eb128016a11fe0eddcc368fb

### DIFF
--- a/.github/workflows/deploy-docs-prod.yaml
+++ b/.github/workflows/deploy-docs-prod.yaml
@@ -18,17 +18,27 @@ jobs:
       - name: Check out repository
         uses: actions/checkout@v2
 
+      - name: "Checkout documentation repository"
+        uses: actions/checkout@v4
+        with:
+          repository: validmind/installation
+          path: site/_source/installation
+          token: ${{ secrets.INSTALLATION_RO_PAT }}
+
       - name: Set up Python 3.11
         uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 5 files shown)


nrichers (changes requested):
@nibalizer this looks to work as expected based on testing but I would ask that you also update `Makefile` for the following actions that it contains the equivalent functionality:

- `deploy-prod`
- `deploy-demo-branch`

(Alternatively, we could chat about removing any deploy functionality from `Makefile` but we've always maintained a local fallback method for deploying.)

EDIT: Can you also include at least a basic description of what this PR does in the root comment, please? 🙂

nrichers (approved):
LGTM! :shipit: 

What I tested: 

- Locally, changes for: 
   - `make get-source` — cleans, clones, and copies as expected, allowing you to bring 
up a preview site with the installation content
   -  `deploy-prod` and `deploy-staging` (without actually deploying) — they copy files as expected 

- On GitHub, changes for: 
   - `validate-docs-site.yaml` — PR preview includes the installation files 

The only thing I didn't test is if `deploy-docs-prod.yaml` will work but the changes look fine.

PR #678, opened 2025-03-20
Remove defunct deploy-demo make action

nrichers:

This PR removes one of those 'we should fix this some day' annoyances that I've spent more time explaining in the last week than it would take to fix. This PR fixes THE LAST PIECE in the puzzle. 

**Before**
We had an extra docs site in the root of the `s3://docs-ci-cd-demo` bucket that goes back to the very early days of Andres and I experimenting with docs deployment. This site serves no purpose, is not current, and is not used — other than we need in index.html in the root of `origin/` for the PR previews to work. 

You could deploy this site with `make deploy-demo`. 

**After**
This site has now been deleted, replaced with an index.html that redirects to https://docs-staging.validmind.ai/ if you try to access the root of `origin/`. Eventually, we might want to list all the PR previews but for now, not having a stale, defunct docs site kicking around that I have to explain is an improvement. 

The `make deploy-demo` is removed in this PR, preventing accidental redeployment of the site.

### DIFF
--- a/site/Makefile
+++ b/site/Makefile
@@ -9,7 +9,7 @@ PROFILE := exe-demo
 FILE_PATH := notebooks/tutorials/intro_for_model_developers_EXECUTED.ipynb 
 
 # Define .PHONY target for help section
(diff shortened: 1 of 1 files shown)


validbeck (approved):
Let's gooo 🎉

nibalizer (approved):

PR #684, opened 2025-04-02
Update the docs site 

nrichers:

This PR includes the source from the `v2.8.13` release tag to refresh our docs site on `prod` and unblock shipping our configurable Docker image.


nrichers:
This PR appears not to be needed — we can just merge `staging` into `prod` and we should be fine.

PR #693, opened 2025-04-17
fix: [sc-9632] Add missing "Export ongoing monitoring document" section to docs

nrichers:
This PR adds the missing steps for exporting ongoing monitoring documentation:

Capto_ 2025-04-16_09-55-43_pm

Story details: https://app.shortcut.com/validmind/story/9632

### DIFF
--- a/site/guide/reporting/export-documentation.qmd
+++ b/site/guide/reporting/export-documentation.qmd
@@ -67,18 +67,16 @@ Export your model documentation, validation reports, and ongoing monitoring docu
 
 ## What's next
 
-* [Working with model documentation](working-with-model-documentation.qmd)
-* [Collaborate on documentation](collaborate-with-others.qmd)
-* [Submit for approval](submit-for-approval.qmd)
-
-
+* [Working with model documentation](/guide/model-documentation/working-with-model-documentation.qmd)
+* [Collaborate on documentation](/guide/model-documentation/collaborate-with-others.qmd)
+* [Submit for approval](/guide/model-documentation/submit-for-approval.qmd)
 
 
 <!-- FOOTNOTES -->
 
 [^1]: [Register models in the inventory](/guide/model-inventory/register-models-in-inventory.qmd)
 
-[^2]: [Working with model documentation](working-with-model-documentation.qmd)
+[^2]: [Working with model documentation](/guide/model-documentation/working-with-model-documentation.qmd)
 
 [^3]: [Manage permissions](/guide/configuration/manage-permissions.qmd)
 
--- a/site/guide/reporting/export-documentation.qmd
+++ b/site/guide/reporting/export-documentation.qmd
@@ -3,6 +3,7 @@ title: "Export documentation"
 date: last-modified
 aliases:
   - /guide/export-documentation.html
+  - /guide/model-documentation/export-documentation.html
 ---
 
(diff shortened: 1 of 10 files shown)


validbeck (changes requested):
The instructions look good, but since we have `Export documentation` as a sidebar entry in both "Model documentation" and "Model validation" now that the ongoing monitoring reports support this let's add this to that section as well before merging. 🙏🏻

nrichers:
> ... since we have `Export documentation` as a sidebar entry in both "Model documentation" and "Model validation" now that the ongoing monitoring reports support this let's add this to that section as well before merging. 🙏🏻

Added and also fixed the listing page issue we spotted during our standup:

Capto_ 2025-04-23_05-40-38_pm

**Note:** There was a link to exporting under the "Reporting" section but when I tested in the UI, no export option is present and our exporting documentation doesn't cover exporting analytics or reports, so I removed the link.

Capto_ 2025-04-23_05-50-25_pm

validbeck:
> Added and also fixed the listing page issue we spotted during our standup:

Not sure why `Export documentation` was removed — it should have both tiles, no?

nrichers:
> I think exporting should stay in that section/nav

Might be faster to discuss for a couple of minutes at our stand-up but is exporting actually supported for analytics? When I checked in the UI, I could not find the option and we don't actually have documentation for exporting analytics, or at least not in the [Export documentation](https://docs-demo.vm.validmind.ai/pr_previews/nrichers/sc-9632/add-missing-export-ongoing-monitoring-document/guide/model-documentation/export-documentation.html#prerequisites) topic that is embedded in the sidebar.

nrichers:
@validbeck in 9e02fbf213697f254de0910d1fc0c9e38ad856ed and 91afcbfebe99dcbe5f63a537feb277bc7fb8e8f8 I tweaked how we embed our `export-documentation.qmd` file:

Capto_ 2025-04-24_04-18-51_pm

(Reporting wasn't shown in the ToC before this commit as it was not at the right heading level, fixed.)

site/guide/reporting/export-documentation.qmd
```diff
@@ -5,7 +5,7 @@ aliases:
   - /guide/export-documentation.html
```
validbeck, line 5:
```suggestion
  - /guide/export-documentation.html
   - /guide/model-documentation/export-documentation.html
```

validbeck, line 5: Missing the alias after our move, but otherwise looks good!

validbeck (approved):
Let's make sure the alias for the old `model-documentation` directory is present, but then it's good to go! 🎉

site/guide/reporting/export-documentation.qmd
nrichers: Committed your suggestion, but please know that somewhere there is a small clock ticking that has just one purpose: to unceremoniouslydocs bite us in the *ss with aliases when we least expect it. Not "if", "when". 😁 (An evil twin problem would be my guess for the most likely issue we hit.)

validbeck: Quarto tells you explicitly when you've used an alias that actually overlaps with an existing, non-broken page — when you preview or render this is caught as part of the output, and in that instance it nullifies the redirect. 

Unless you mean another "evil twin" problem?

PR #695, opened 2025-04-22
Release Notes — Unified version `23.03.05`

validbeck:

> sc-9279

### April 24, 2025

[**LIVE PREVIEW**](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-9279/release-notes-to-sprint-70-25-03/releases/2025/2025-apr-24/release-notes.html)

> [!NOTE]  
> @emmavdh Can you please review the verbiage for the release notes above and make sure the delivery looks good to you? Thank you! 🙏🏻 

#### ValidMind Library version

cc: @nrichers 

The version exposed in the release notes is ValidMind Library `v2.8.13`, but according to @nrichers this version is actually "broken" — the Python API docs weren't gathered up into this release due to a workflow blip and thus can't be pulled in from the source at this release tag.

We decided that the workaround until we catch up is just to pull in the latest production version (as the most recent one doesn't have that many external changes), so while I've hacked the files to say `v2.8.13` this is actually technically `v2.8.17`.

##### Jupyter Notebooks

cc: @nrichers 

As the documentation side references notebooks past the production ValidMind Library versions (the new validation notebooks, mainly), I've reverted the `notebooks/` directory and `notebooks.zip` to what currently exists on `main` to ensure no broken links/gaps in the surfaced documentation.

We will continue to need to do this until we catch up with our versioning, which should be the next release tag that goes out for the library. 

### New "Releases" landing

[**LIVE PREVIEW**](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-9279/release-notes-to-sprint-70-25-03/releases/all-releases.html)

> [!NOTE]  
> @nrichers As this makes changes to our documentation site structure and authoring experience, please review the following items in this section. 🙏🏻 

To accomodate for the homepage redesign that includes a more streamlined releases section, including a link out to "All releases" and indications for the unified version, I've now built a **Releases** landing powered by it's own `_sidebar.yaml`:

| Homepage redesign mockup | New Releases landing | Updated latest release section |
|---|---|---|
|Screenshot 2025-04-22 at 11 32 18 AM | Screenshot 2025-04-22 at 11 32 54 AM|Screenshot 2025-04-22 at 11 33 38 AM |

- This new releases landing includes an all releases listing page, which automatically lists and sorts all releases from all time in descending thanks to one minor adjustment — simply including the date field on each release file.

- This all releases listing operates with wildcards and  **exclusion criteria based on category for the entire releases folder** — this is why the yearly release roundups and the breaking changes/deprecation have special category tags. 

- **The release notes authoring experience does not change** — in fact, it's been simplified — refer to the updated scripts and notebooks below for more information.

##### Generate release notes script & notebook

- These now automatically insert the date field in the correct format just like we insert the human-readable date as the release title.

- Additionally, **these now prompt you to supply a unified version,** which is automatically captured in the release file as the subtitle:
Screenshot 2025-04-22 at 11 37 28 AM

- As we moved the releases into their own landing with their own `_sidebar.yaml`, these now update this file instead of ths main `_quarto.yml`:
https://github.com/validmind/documentation/blob/0385fc0aae9a446d6b60fa16154d6177a97fa872/site/releases/_sidebar.yaml#L3

##### Yearly release cleanup script & notebook

- As we moved the releases into their own landing with their own `_sidebar.yaml`, these also now update this file instead of ths main `_quarto.yml`.

- These used to find and locate all the releases for one year and sort them by date based on file directory name to generate indiviual listing items, but since we now have dates on all release files these listings are simply a wildcard entry sorted by date. 🎉 

#### Makefile cleanup

- There were missing help & `PHONY` entries so I fixed that:
https://github.com/validmind/documentation/blob/0385fc0aae9a446d6b60fa16154d6177a97fa872/site/Makefile#L14

- I also sorted the commands by ABC order when it didn't affect the execution — it bothered me to not be able to easily locate the command I wanted to modify.

#### FAQ cleanup

[**LIVE PREVIEW**](https://docs-demo.vm.validmind.ai/pr_previews/beck/sc-9279/release-notes-to-sprint-70-25-03/faq/faq.html)

I realised after I used the categories to exclude listing items for releases that we could just apply the categories on the main listing to the FAQ individual pages (where edits to information are usually done anyway), which also gives users a summary of what answers are on the page. For example:

Screenshot 2025-04-22 at 11 34 39 AM


site/releases/2025/2025-apr-24/release-notes.qmd
```diff
@@ -0,0 +1,284 @@
...
+::: {.w-20-ns .tc}
+[result.log()](/validmind/validmind/vm_models.qmd#log){.button target="_blank"}
+
+:::
+
+::::
+
+:::: {.flex .flex-wrap .justify-around}
+
+::: {.w-30-ns}
```
nrichers, line 157:
```suggestion
::: {.w-30-ns .pr1}
```

Just adding a bit of padding here and in the right column, see next comment. Tested locally, works without causing text wrapping.

site/releases/2025/2025-apr-24/release-notes.qmd
```diff
@@ -0,0 +1,284 @@
...
+
+- `hideTitle`: Title
+- `hideText`: Description
+- `hideParams`: Parameters
+- `hideTables`: Tables
+- `hideFigures`: Figures
+
+:::
+
+::: {.w-70-ns}
```
nrichers, line 168:
```suggestion
::: {.w-70-ns .pl1}
```

nrichers (approved):
Thank you for this PR, @validbeck! 🚀🚀🚀 

Workarounds for the Python API and notebooks look reasonable — thank you for figuring these out.

The new `_sidebar.yml` also makes sense, acknowledged that this is what we will use going forward.

The landing page for releases is a nice touch, but it feels slightly ahead of where we are with the homepage redesign. Most notably, the Figma mockup includes a short description at the top, and the format differs — combining version and date on a single line. I would not make further changes in this PR but we should like chat about how we can include content that will work with the new design and then have a consistent design on this new landing page for release notes.

Capto_ 2025-04-22_02-26-54_pm

I assume the release notes for the new model validation notebooks will appear in a future release? I guess it's the lag between merging something and seeing it in the notes again — ties into the broader discussion about the release notes process that came out of our offsite.

validbeck:
> The landing page for releases is a nice touch, but it feels slightly ahead of where we are with the homepage redesign. Most notably, the Figma mockup includes a short description at the top, and the format differs — combining version and date on a single line. I would not make further changes in this PR but we should like chat about how we can include content that will work with the new design and then have a consistent design on this new landing page for release notes.

This is just setup for the new landing, not the "final" product — we can very easily massage the CSS for the release listing tiles to display the title & subtitle on the same line, for example! The information wasn't there before, and it is now — and this also allows us to easily display the unified versioning on releases going forward without the bold callout as this becomes the norm.

But I don't think we should match the homepage tile to the actual release page — they are different formats. As long as the same information is there, I think it is fine. 

> I assume the release notes for the new model validation notebooks will appear in a future release? I guess it's the lag between merging something and seeing it in the notes again — ties into the broader discussion about the release notes process that came out of our offsite.

Yes, they are not actually out in the latest release tag for the `validmind-library` — hopefully the release tags will catch up with the material we have out soon and we can stop cherry-picking the notebooks.

site/releases/2025/2025-apr-24/release-notes.qmd
```diff
@@ -0,0 +1,284 @@
+---
+title: "April 24, 2025"
+date: 2025-04-24
+subtitle: "Unified version `23.03.05`"
```
nrichers, line 4: Just noticed this version: that should likely be `25.03.05`?

validbeck: Ah, good catch, I was confused by this typo from Spencer: https://validmind.slack.com/archives/C087ES2CD1B/p1743025690470939

emmavdh (dismissed):
Looks really great! One minor thing - line 228, this line seems clunky/confusing: 
"Previously, when test results logged to the {{< var validmind.platform >}} contained more than 20 pages of data in a table were inserted as test-driven blocks into documentation, the previous and next buttons did not render correctly for pagination."

Maybe something like this?
"Previously, when test results logged to the {{< var validmind.platform >}} contained more than 20 pages of data in a table, the previous and next buttons did not render correctly for pagination when added as test-driven blocks into documentation."

validbeck:
> Maybe something like this?
"Previously, when test results logged to the {{< var validmind.platform >}} contained more than 20 pages of data in a table, the previous and next buttons did not render correctly for pagination when added as test-driven blocks into documentation."

Thanks @emmavdh, this is a great suggestion — I 100% struggled with the wording in that section so this is very helpful. 💖 

I've made the changes directly, but you can in the future suggest them as part of your review (it means you get ✨ credit ✨ for the edit/become part of the commit history!): https://www.youtube.com/watch?v=ciJpIWs3t3k

PR #703, opened 2025-04-28
fix: remove helm repo path from s3 sync commands

eggshell:
This path stores our helm chart tarballs, and objects in that path should not be removed when the s3 sync command runs.


nrichers (dismissed):
These changes look fine, but it feels like the fact that we copy these files into some AWS S3 buckets should be documented somewhere? I didn’t know we did this — I’m sure there are other things we’ll need to watch for (looking at you, [#704](https://github.com/validmind/documentation/pull/704) ...).

At a minumum, there is [Docs CI/CD Deployment](https://www.notion.so/validmind/Community-Processes-b33e9e56c3b2461796ee0f71697132bf?p=9691830a10d14a3f963daafab278e25b&pm=c) — please document your changes there?

nrichers:
@eggshell we also needed to update our Makefile in case we deploy manually, see https://github.com/validmind/documentation/pull/703/commits/fb5ded505b711664933a7aedd03345765a53a692. 

(Talked to @nibalizer about us avoiding all this parallel stuff last week, ideally we should agree on a central place like a Makefile that all the other places use.)

nibalizer (approved):

nibalizer (approved):

PR #854, opened 2025-07-22
PR preview for simple IDP Auth docs / basic ValidMind user role info

nrichers:
# Pull Request Description

## What and why?

This PR:

- Provides a preview for the restructured authentication user guide content, including simple IDP auth.
- Includes a new section on basic ValidMind user access

Rel. [sc-11173](https://app.shortcut.com/validmind/story/11173/documentation-for-simple-idp-auth-mode)

## How to test

This is the new section on basic ValidMind user access: [Basic ValidMind access](https://docs-staging.validmind.ai/pr_previews/nrichers/sc-11173/documentation-for-simple-idp-auth-mode/guide/configuration/log-in-to-validmind.html#basic-validmind-access)

Capto_ 2025-07-22_05-36-26_pm

IMPORTANT: Remove the `ref` from the workflow before merging (mainly a note for Nik)

## Checklist
- [x] What and why
- [x] How to test
- [x] Dependencies, breaking changes, and deployment notes
- [x] Labels applied
- [x] PR linked to Shortcut
- [ ] Tested locally
- [x] Documentation updated (if required)

### DIFF
--- a/.github/workflows/validate-docs-site.yaml
+++ b/.github/workflows/validate-docs-site.yaml
@@ -22,7 +22,6 @@ jobs:
       with:
         repository: validmind/installation
         path: site/_source/installation
-        ref: nrichers/sc-11173/documentation-for-simple-idp-auth-mode
         token: ${{ secrets.INSTALLATION_RO_PAT }}
 
     - name: Check out release-notes repository


nibalizer (approved):

site/guide/configuration/log-in-to-validmind.qmd
```diff
@@ -53,10 +53,22 @@ Log in to our cloud-hosted {{< var validmind.platform >}} to work with your mode
 
     After successful login, you will be redirected to the main dashboard of the {{< var validmind.platform >}}.
 
+## Basic ValidMind access
+
+If you are a new user signing in for the first time, you might have only _a basic ValidMind user role_. This role allows you to authenticate but provides limited access to your organization, models, and platform features.
```
jamadriz, line 58: @nrichers we have two possible flows for user signing in for the first time -  this determined by a global Super Admin Setting for the site - 
1. allow user self onboarding. If this is enabled, user can sign up, and two flows can happen here:
1.1 user has been invited to an org, in this case, invitation will be accepted automatically and logged in to the organiation (with the roles coming from the invite)
1.2 user doesn't have an invite nor org, in this case, the user can create an org, and will be assigned as Customer Admin / Developer by default to that org
2. if user self onboarding is disabled, a page will be displayed indicating user to contact admin to be assigned to an Org

nrichers: @jamadriz thank you for this review comment! I made some updates in e284aba31. 

This is what it now looks like, let me know if further changes are needed? 

Capto_ 2025-07-24_05-29-06_pm

jamadriz: Not sure if its relevant, but `Basic User` is still tied to an Organization, so the text is quite not correct. The initial intention was what you described, but there is nothing to do outside of an Organization context (perhaps changing profile but even that can be restricted by IDP), for a ValidMind non Admin user, and having a role outside of an Organization context didn't make sense. 

A `Basic User` role intention now covers a gap we've had for some time, were the initial organization roles of Validator, Developer might not be applicable to all users, and also that if a user belongs to an Org, we want to ensure it always has a set of default permissions in the App (under Organization Context). And eventually we might even want to rethink Developer, and Validator as Organization level roles since they make more sense at a Stakeholder Type context.

nrichers: > the message is to wait for an admin to assign to org

OK, I pushed 174ed55ad to make one more change, but I kept it very general ("... administrator must provide you with the necessary access ..."). I hope that works for you.

.github/workflows/validate-docs-site.yaml
```diff
@@ -22,6 +22,7 @@ jobs:
       with:
         repository: validmind/installation
         path: site/_source/installation
+        ref: nrichers/sc-11173/documentation-for-simple-idp-auth-mode
```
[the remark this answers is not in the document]
nrichers: Removed temporary branch `ref` in 2a6b69fbc

PR #921, opened 2025-08-29
Removing validmind version variable

validbeck:
# Pull Request Description

## What and why?

> sc-7174

We now pull in the production library version in the Python API docs, and don't reference the library version elsewhere. Removing the variable to reduce confusion/accidental usage.

(I considered populating the variable in the `_variables.yml` file with a script from the source, but we don't even use the variable anywhere for that much work to be put into that workaround . . .

## How to test

n/a

## What needs special review?

n/a

n/a

## Release notes

n/a

## Checklist
- [x] What and why
- [ ] Screenshots or videos (Frontend)
- [ ] How to test
- [x] Labels applied
- [x] PR linked to Shortcut
- [x] Tested locally

### DIFF
--- a/site/_variables.yml
+++ b/site/_variables.yml
@@ -79,7 +79,7 @@ gcp:
 
 version:
   python: "≧3.8 and <3.11"
-  validmind: v2.5.19
+  # validmind: v2.5.19
 
 # OTHER
 


nrichers (approved):
LGTM, but I would recommend just removing the line since we've never used it.

PR #949, opened 2025-09-17
Replace Log In menu with button

nrichers:
# Pull Request Description

## PR Summary: Convert Login Dropdown Menu to Button

This PR changes the login option in our docs site from a dropdown menu to a direct button link and reorganizes some CSS to make it easier to maintain.

Relates to [sc-12287](https://app.shortcut.com/validmind/story/12287/change-login-menu-on-docs-site-to-be-a-direct-login-link)

### Quarto profile changes

I removed the dropdown menu and replaced it with a simple link that points to the correct location for each profile:

- `_quarto-development.yml` points to https://app.dev.vm.validmind.ai/
- `_quarto-staging.yml` points to https://app.staging.validmind.ai/
- `_quarto-production.yml` points to https://app.prod.validmind.ai/

### CSS / SCSS changes

Some CSS changes are required because Quarto's navbar YAML doesn't support adding CSS classes directly to individual nav items. In the spirit of improving our CSS rather than adding to the jungle, I reorganized some CSS files and switched to SCSS: 

  -  `scss/_buttons.scss` - Contains all button styles (`.button`, `.button-green`, `.cta`, `#nav-menu-log-in`)
  - `scss/_links.scss` - Contains all link styles (nav-links, sidebar-links, footer-links, TOC links)
- `scss/_termsfeed.scss` - Some random CSS at the bottom of our `themes.scss` that I can't recall what it does,  moved it into it's own file while adding the `@import` statements nearby
- `scss/_codeblocks.scss`, `site/scss/_bug-fixes.scss` - Moved to the SCSS folder and renamed 
- Updated `theme.scss` to import new modular files
- Cleaned up `styles.css` by removing button and link styles

To apply the button styling to the login link:

- Used `@extend #nav-menu-log-in` to style the login nav-link using existing button styling (yay!)
- Added `text-decoration: none !important` on hover to prevent underlines on the button link
- Added some additional horizontal padding to ensure the Log In link looks correct

## How to test

You can check out the PR and test locally with `--profile` to see the difference in the direct login URLs. 

### Before

image

### After, `--profile development`

Capto_ 2025-09-17_11-42-43_am

Capto_ 2025-09-17_11-45-05_am

### After, `--profile staging`

Capto_ 2025-09-17_11-42-43_am

Capto_ 2025-09-17_11-49-38_am

### After, `--profile production`

Capto_ 2025-09-17_11-42-43_am

Capto_ 2025-09-17_11-43-00_am

## Release notes

[...]

PR #956, opened 2025-09-19
Improve navigation for Python API and Public REST API — DO NOT MERGE

nrichers:
# Pull Request Description

## What and why?

In this PR:

- **Footer and navigation dropdown** — Added API reference links to Public REST API and Python API
- **Footer layout** — Reduced the "Guides" section from three to two columns and reshuffled content
- **Refactor CSS** — Moved footer-specific styles into a dedicated `_footer.scss` file for better maintainability

Notes:

- Navbar CSS: I did not modify the navbar CSS because some class names closely resemble those used for the sidebar. To keep the refactor simple and safe, I focused only on the footer.
- Footer layout changes: Adding the Public REST API link to the “Library” section makes that column noticeably taller than the others. Reducing the “Guides” columns from three to two creates a more balanced layout and lowers the minimum window width required to display the full footer. 

Relates to [sc-12329](https://app.shortcut.com/validmind/story/12329/add-public-api-to-footer); blocked by #949 

## How to test

Take a look at the preview or compare the screenshots: [**LIVE PREVIEW**](https://docs-staging.validmind.ai/pr_previews/nrichers/sc-12329/add-public-api-to-footer/index.html)

### Before 

Capto_ 2025-09-19_11-09-20_am

### After

Capto_ 2025-09-19_11-12-22_am

## Checklist
- [x] What and why
- [x] How to test
- [x] Labels applied
- [x] PR linked to Shortcut
- [x] Tested locally


validbeck:
Going to take a closer look, but I don't like this:

> Footer layout — Reduced the "Guides" section from three to two columns and reshuffled content

I massaged the footer into the three columns initially because the two columns looked un-equal and cramped. Can we please revert? Your columns are actually also missing a link! (Access)

site/_quarto.yml
```diff
@@ -49,9 +49,13 @@ website:
         - text: "{{< fa bullhorn >}} Releases"
           file: releases/all-releases.qmd
         - text: "---"
-        - text:  "{{< fa cube >}} Python Library"
+        - text:  "{{< fa cube >}} Library"
```
validbeck, line 52:
```suggestion
        - text:  "{{< fa cube >}} Reference"
```

validbeck, line 52: I think "Reference" fits better here to reduce confusion, since we have a product we call the library. There are references under here that don't cover our library product. ;)

validbeck (changes requested):

site/_quarto.yml
nrichers: As discussed, I removed the labels & icons.

nrichers:
> I massaged the footer into the three columns initially because the two columns looked un-equal and cramped. Can we please revert? Your columns are actually also missing a link! (Access)

As discussed, I re-added the access "Access" and balanced the links across the columns. 

FYI, there are now some column spacing issues, I have a CSS fix in the works that gets them looking like this and wraps correctly, will tag you for re-review when I have completed testing:

image

nrichers:
> The top and bottom padding seems a little squished as well, I tried to poke around in it but admittedly I haven't spent much time with the footer so I'm not sure if I'm looking in the right place... 🤔

I would hold off on poking around; I'm four hours deep into the footer CSS. Might be easier for you to just review my changes when they are ready.

nrichers:
OK, many hours later ...

- Footer display should be fixed on desktop, responsive, and mobile views for the most part (see below).  
- The main `themes.css` now imports only other SCSS files, cleaning up where we define our SCSS styles. (The import order matters, FYI.)
- Many colors are now defined once in `_validmind_colors.scss` and reused as variables elsewhere.  
- New SCSS files continue the process of pulling CSS out of `styles.css` and moving existing SCSS into better locations (examples: `_dropdowns.scss` for dropdown menus; `_footer.scss` now contains all footer definitions instead of defining footer links in `_links.scss`).  
- Some duplicated CSS made testing time consuming. I’ve removed what I found, but more cleanup is to come. 

Overall, other than the footer fixes, the CSS should behave as before, but I would welcome some independent review to make sure I have not missing something. I'm at the 'not seeing the forest for all the trees' stage with this PR and I'm tired testing. 🌲🌲🌲 

## Footer display

### Desktop mode (1440x900)

image

### Responsive mode

Starting to wrap:

image

The most we will wrap before hiding the links:

image

**Issue**: I would prefer not to wrap the links section (`.nav-footer-center`) into more than two rows before hiding. I tried several methods, but there continues to be a conflict between our styles and Bootstrap. For now, I’ve left it at improving the specificity of our footer styles. Other attempts that made no difference include:

- Using even stronger specificity and more comprehensive hiding rules
- Applying specific Bootstrap override instructions (`@layer bootstrap-override`)  

If you have some clever ideas, I'd love to hear about them. For now, I am out of time on investigating this further. 

Small screens: 

image

### Mobile mode

image

nrichers:
@validbeck one other thing before I re-request review from you: there are some specificity differences when you use SCSS versus `styles.css`. My take is that some differences are OK, but if there are notable ones you think we should fix, let me know (as I am sure you will!).

validbeck (changes requested):
The links still don't quite look right on hover:

| OLD | CURRENT |
|--|--|
| Screenshot 2025-09-25 at 11 39 03 AM | Screenshot 2025-09-25 at 11 39 16 AM|

In this case I personally *like* the new non-underlined link better, but having an underline *somewhere* in the link interaction process is a good accessibility tool . . .

Footer has all the stuff (I think) but the top and bottom padding is still really squished. For example:

| OLD | CURRENT |
|--|--|
| Screenshot 2025-09-25 at 11 37 45 AM | Screenshot 2025-09-25 at 11 41 18 AM |

I tried playing around with it locally but nothing I did to the padding seems to have adjusted anything, so I'm not sure how that's all working.

nrichers:
> Footer has all the stuff (I think) but the top and bottom padding is still really squished.

This is fixed in 0b1007dfc:

image

> The links still don't quite look right on hover

This should be fixed in 78b5e8ce0 and 175c4e832:

https://github.com/user-attachments/assets/cb7ff71c-a2bb-4398-b65e-4290bae41683

With all of these issues, the correct approach that gives most fine-grained control seems to be:

- Find the right CSS selector, the more specific the better
- Search for any duplicated CSS that might be causing the issue and resolve
- Determine if it should be a _SCSS default_ that overrides Bootstrap or a _SCSS rule_ (note SCSS defaults MUST live in `themes.scss`, fixed in 20e1f5ce8)
- If our styles still get overriden by Bootstrap, typically because they are a rule and not a default, _selectively_ apply `!important`

PR #966, opened 2025-09-25
Basic version: Improve navigation for Python API and Public REST API

nrichers:
# Pull Request Description

## What and why?

This PR applies only the following changes from #956: 

- Fix links to Python API and Public REST API in footer & dropdown
- Remove icons and text labels from dropdown
- Fix link color and remove underline

Relates to [sc-12329](https://app.shortcut.com/validmind/story/12329/add-public-api-to-footer)

Capto_ 2025-09-25_03-55-24_pm

Capto_ 2025-09-25_04-18-27_pm

To get this PR here into `prod` safely, I will cherry pick the merge commit only.

## Checklist
- [x] What and why
- [ ] How to test
- [x] Labels applied
- [x] PR linked to Shortcut
- [x] Tested locally

### DIFF
--- a/site/assets/styles/_links.scss
+++ b/site/assets/styles/_links.scss
@@ -7,11 +7,13 @@ a {
 }
 
 /* Section-specific underline treatments */
-#prerequisites li > label a:hover,
-#whats-next ul li a:hover,
-div[id^="fn"] > p > a:hover,
-#submit-feedback ol li p a:hover,
-#connect-with-us ul li > a:hover {
+div[id^="fn"] a:hover,
+#connect-with-us a:hover,
+#prerequisites a:hover,
+#supported-model-types a:hover,
+#supported-modeling-libraries-and-other-tools a:hover,
+#submit-feedback a:hover,
+#whats-next a:hover:not(.quarto-grid-link) {
   text-decoration: underline 2px solid #083E44 !important;
 }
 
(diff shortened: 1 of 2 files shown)


validbeck (changes requested):
In this case I think we should actually retain the link changes from https://github.com/validmind/documentation/pull/949. I like the no-underline at all even on hover better personally, but it's just not very accessible.

nrichers:
> In this case I think we should actually retain the link changes from https://github.com/validmind/documentation/pull/949.

@validbeck I ended up adding some very specific SCSS to underline only specific types of links. 

https://github.com/user-attachments/assets/127d438f-6db0-442c-a601-89fbe2324412

nrichers:
@validbeck final commit on this PR from me, this one 

- fixes the issue with underlines on links that are inside of <ul>
- prevents underlines on listing cards, required because some of our what's next links are basic links and some are cards.

If this is not acceptable, then I will close this PR - most of the changes I've made over the last couple of days are stop-gap measures that should go away once we sort out more of our SCSS correctly. 

(For example, it would be much simpler to specify a single rule on underlines for links that includes a directive about which  which links we DON'T want to underline globally. But this PR doesn't include some of the improvements from the other PR that are required to implement this cleanly and there's no point retrofitting them if the work's already done elsewhere.)

PR #1288, opened 2026-04-23
docs: Enabling org-wide track changes disables AI-content drafts

validbeck:
# Pull Request Description

## What and why?

> sc-15852

This has come up in a Support ticket and I noticed this quirk while testing — this is behaviour that is working as intended so I've documented it.

## How to test

Click on the live previews and compare the changes:

| page | OLD | NEW |
|--|--|--|
| [Managing your organization](https://docs-staging.validmind.ai/pr_previews/beck/sc-15852/documentation-track-changes-by-default-disables/guide/configuration/managing-your-organization.html#manage-document-defaults) |Screenshot 2026-04-23 at 2 16 19 PM |Screenshot 2026-04-23 at 2 16 44 PM |
| [Work with content blocks](https://docs-staging.validmind.ai/pr_previews/beck/sc-15852/documentation-track-changes-by-default-disables/guide/model-documentation/work-with-content-blocks.html#generate-content) |Screenshot 2026-04-23 at 2 16 35 PM |Screenshot 2026-04-23 at 2 16 51 PM |

## What needs special review?

n/a

n/a

## Release notes

n/a

## Checklist
- [x] What and why
- [x] How to test
- [x] Labels applied
- [x] PR linked to Shortcut
- [x] Tested locally
- [x] Documentation updated (if required)

### DIFF
--- a/site/guide/configuration/managing-your-organization.qmd
+++ b/site/guide/configuration/managing-your-organization.qmd
@@ -71,7 +71,10 @@ By default, the [{{< fa hand >}} Customer Admin]{.bubble} role has these permiss
 ### Tracked changes
 
 ::: {.callout-important}
-Enabling tracked changes by default for your organization will disable content draft generation with AI.[^5]
+Enabling tracked changes by default for your organization will disable the following within simple text blocks:[^5]
+
+- The ability to reference field values in the form of variables
+- Content draft generation with AI
 
 :::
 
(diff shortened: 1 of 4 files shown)


cachafla (approved):
Looks great, although we might also need to say that the "Insert Variables" feature is also affected. The reason these features break is that they are custom plugins and CKEditor forcefully disables plugins that have not built track changes support in their APIs, since it wouldn't know how to compute the before/after states when using them.

This means:
1) Any new plugin we make will be affected
2) But we can fix this - just need to look at how the track changes API works. This is low priority but not out of consideration

validbeck:
> Looks great, although we might also need to say that the "Insert Variables" feature is also affected. The reason these features break is that they are custom plugins and CKEditor forcefully disables plugins that have not built track changes support in their APIs, since it wouldn't know how to compute the before/after states when using them.

@cachafla Oh interesting, I see that now. Made an update:

| managing your organization | work with content blocks |
|--|--|
|Screenshot 2026-04-24 at 9 01 19 AM |Screenshot 2026-04-24 at 9 16 50 AM | 

We also removed the ability to insert variables into text blocks, so I hid the info for now:

| old | new |
|--|--|
|Screenshot 2026-04-24 at 9 12 34 AM |Screenshot 2026-04-24 at 9 16 23 AM |

